PR #60, opened 2020-05-26
bFn changed to match with buffer.js prototype

fsamuelm:
"readDoubleBE" has to be just "DoubleBE", to comply with buffer.js.
At runtime, concatenation with the prefix "read" is already being done, which was resulting in "readreadDoubleBE"


unocelli:
Hi Fábio, Thanks!

PR #67, opened 2020-08-04
Update copyright year

Joelius300:
I updated the copyright year in the license file to the current year and made it a range.

This is the very simple PR we talked about earlier ;)


unocelli:
Thanks Joel

Issue #92, opened 2021-01-26
Proposal. Add views templates for cards/dialogs


NikolayMurha:
Hello.

I have dashboard with >40 semaphores and I need dialog for each of them.
All dialogs identical, but variables different for each dialog. 
Now I create 40 different views for each dialog.
I think, will be good to make views templates with placeholders instead of variables names, and set variables names on dialog properties.

unocelli:
I think the issue is solved. I'm therefore going to close it.

Issue #90, opened 2021-01-19
Missing #blur-filter mask on home page


NikolayMurha:
Hello.

Filter #blur-filter exists on editor but not on home page: 
https://github.com/frangoteam/FUXA/blob/master/client/src/app/editor/editor.component.ts#L430
As a result, gauge-semaphore don't work correctly.
Mac OS, Chrome.

unocelli:
Hi,
can you better explain what is not working correctly?
The Filter #blur-filter should be added to every view, in the home page too. Have you inspect the DOM elements (the first <svg> in <app-fuxa-view>)?

NikolayMurha:
When I add semaphore it's disappeared when gauges gauge condition is true, but should change color instead.
I investigated the problem and came to the conclusion that the filter element is missing.

Removing the line solves the problem: https://github.com/frangoteam/FUXA/blob/master/client/src/app/gauges/controls/gauge-semaphore/gauge-semaphore.component.ts#L53 
I think this is dirty solution, but I didn't figure out where the blur-filter gone.

[Screenshot 1](https://www.dropbox.com/s/zn78mmgs1xobm43/%D0%A1%D0%BD%D0%B8%D0%BC%D0%BE%D0%BA%20%D1%8D%D0%BA%D1%80%D0%B0%D0%BD%D0%B0%202021-01-22%20%D0%B2%2017.38.31.png?dl=0)
[Screenshot 2](https://www.dropbox.com/s/sj5eexg9tz5nmd5/%D0%A1%D0%BD%D0%B8%D0%BC%D0%BE%D0%BA%20%D1%8D%D0%BA%D1%80%D0%B0%D0%BD%D0%B0%202021-01-22%20%D0%B2%2017.39.12.png?dl=0)

unocelli:
OK, I have to investigate why by Mac OS, Chrome don't work. 
The question is also whether the blur effect is interesting?
![screenshot](https://user-images.githubusercontent.com/48684550/105516448-2fd24c00-5cd6-11eb-80d5-53dfe5910a4c.PNG)

NikolayMurha:
No, not interesting for me :)

unocelli:
I'm agree with you, the blur effect is not so sexy

NikolayMurha:
I would have leave it, but I could not figure out why this element missing on mac. Semaphore without blur better than transparent semaphore :)

PR #94, opened 2021-01-31
Add ability to edit variables names manually and override them on card/dialog 

NikolayMurha:
Hello. This pull request is an evolution of my proposal #92. 

- Add posability to set custom variable names and override it on dialog or card settings. It's add posability to create views as templates for cards/dialogs  #92. It also allows  to set a static value for variables.

Along the way:

- Remove bur filter #90
- Add Roboto cyrillic fonts
- Revert workarea scrollbars. With shift key zoom it's loks better (on mac)


NikolayMurha:
I will split this pull request

PR #96, opened 2021-02-01
Revert scrollbars for workarea

NikolayMurha:
I propose to revert scrollbars for workarea element. It's very useful on mac with shift+scroll zoom.


unocelli:
why not... ;)
I have also to fix the color palette horizontal scrollbar

PR #97, opened 2021-02-01
Add posability for set properties for any shapes.

NikolayMurha:
I try to add posability to set actions for any shapes. This made it possible to create complex elements that are not in the library. 
For example I created switch shape (group) which contains two other shapes (on and off) plased stacked with reverted show action. And this work good


unocelli:
Sounds good...show me when it's ready

PR #99, opened 2021-02-01
Add feature for set custom variables names/values and override it on …

NikolayMurha:
Hello. 

This pull request is an evolution of my proposal #92.
It add posability to set custom variable names and override it on dialog or card settings. It's add posability to create views as templates for cards/dialogs #92. It also allows to set a static value for variables.


unocelli:
I will have to do some tests with it before merging.

unocelli:
Hi Nikolay,
Really a great job! Bravo! Thank a lot for the code refactoring.
I wonder if it wouldn't be better to show all variables by default instead of having to add each one?

PR #100, opened 2021-02-02
Add ranges colors for buttons and ToggleValue event type

NikolayMurha:
Hello.

I have add ranges property for button background color and Toggle Value action for events.

Now you can create toggle buttons with color indication


unocelli:
Hi,
for future reference, I think you should first open an issue explaining the problem but you've done a great job at the PR description so you don't need to create one now :)

I'm making a contributing guidelines and I know my code is not always very consistent ;), but I would appreciate if we could follow the indent with 4-spaces, no tabs :)

Thanks a lot for your help

NikolayMurha:
Hello. 
Thank you for the comment. 

Now I am setting up some dashboard, and when I run into a limitation, I try to quickly solve it. Along the way, I offer my work to you. I am not a great specialist in Angular and Typeescript. Before that I used vue.js in my projects with vanila js, so comments are necessary. Anyway, I gave you the code and the final decision is yours.

Also, I wanted to raise a formatting issue, but you got ahead of me :) My editor is using your .editorconfig and there is a 2-space indent, so I think when yoy will edit .editorconfig aloung your codding styles. This will helps all of us syncronize code style.
Thanks.

unocelli:
Good job. Thanks

Issue #102, opened 2021-02-05
Cannot read property 'numberOfItems' of undefined


NikolayMurha:
Hello.

After resizing some grouped objects I have error on save:

`core.js:1673 ERROR TypeError: Cannot read property 'numberOfItems' of undefined`
 

Apparently, this is a rare error, but I have it.

Step for reproduce:
1. Load project https://gist.github.com/NikolayMurha/ceb0e5d365c011089d5a6f759bbc8e60
2. Save and reload (just in case)
3. Resize green object
4. Try to save

P.S. Pull request soon

unocelli:
Great job Nikolay, I hope to see a demo of your shapes ;)

PR #103, opened 2021-02-05
Fix "Cannot read property 'numberOfItems' of undefined" error 

NikolayMurha:
Hello.

This workaround for issue #102

I had to partially de obfuscate your [fuxa-editor.js](https://github.com/NikolayMurha/FUXA/blob/master/client/src/assets/lib/svgeditor/fuxa-editor.js) to understand what the problem is. That was not easy ;) Unfortunately I didn't find the root of the problem, but I found a solution on [svg-edit repo](https://github.com/SVG-Edit/svgedit/blob/2235f81ee7e3bf927126de10de03e7ecdd396f60/editor/utilities.js#L1240).

The essence of the problem is that the `line` element gets the `d` attribute `null` value and `line` elements process as a svg path, but it is not a path and has no properties.
this fix skip null atributes.

I specially formatted the piece of code so you can see what I changed.


unocelli:
Great job, Thanks

Issue #78, opened 2020-10-30
Property window confirm buttons outside browser window


RG8550:
Can not confirm selected tags in Tag browser because button outside browser window.
![tag browser no button](https://user-images.githubusercontent.com/73709550/97726556-9ed14780-1acf-11eb-984a-0aa6dad14762.jpg)

For every object property window are the buttons outside browser window.
Can not scroll down to buttons.
![input no buttons](https://user-images.githubusercontent.com/73709550/97728598-08525580-1ad2-11eb-9cd9-75ec7a0a4fed.jpg)
![gauge no button](https://user-images.githubusercontent.com/73709550/97727269-844b9e00-1ad0-11eb-95b9-9fde1a574718.jpg)

Browser used: Chrome 86.0.4240.111

This is on a fresh install of ubuntu server 20.04.
nvm 0.34.0
node 10.17.0
npm 6.11.3
python 2.7.18

Nothing changed in settings.js.

No warnings/errors in fuxa.log
Only noticeable in api.log:
`"GET /assets/lib/gauge/gauge.js.map HTTP/1.1" 404 168 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/86.0.4240.111 Safari/537.36"`

unocelli:
Hi,
tanks for the feedback, I'll see if I can activate the scrollbar but I think to work comfortably with a graphics editor it is better to use a fairly large display and you know you can easily zoom with browser (CTRL +/-)

RG8550:
Thanks for looking to get scrollbar working.

Yes buttons are visible when zooming out browser.
But browser zoom does not work when in views editor, need to go to different page to zoom out.

unocelli:
I'm going to close this as resolved. let me know if you have any issues.

PR #104, opened 2021-02-05
Use relative units for gauge property dialog

NikolayMurha:
In this PR I try to fix issue with gauge property dialog height #78 and make active toolbar button highlighted on hover.


unocelli:
Hi,
I'm sorry for my delay, I tried to solve the conflict and I hope It's ok...Thanks

PR #140, opened 2021-05-08
Fix #139 - use secure fields (type="password") for password input.

### LINKED ISSUE #139: Password fields are plain text so that passwords are always visible when entered.
image

and 

image

tromgy:
image

and

image

and also fix a typo.


unocelli:
Good job. Thanks...You don't think that now we need the show password button

PR #163, opened 2021-06-09
Only Change is OPC UA Variable name usage

alpaslanske:
I changed variable name usage of OPC UA. Changes is only visual. 
When adding tags you will see varable name of tag taken from OPC UA server.
When assigning tags to controls  you will see variable name of tag taken from OPC UA server.

Before these changes you saw names like n=4;i=7 which is not meaningful to humans.


unocelli:
I do some tests with it before merging.

unocelli:
Hi,
Good job, I removed the file created from visual studio and I added the partial turkish translation, I hope you are not angry ;)

PR #172, opened 2021-06-27
Full Turkish translation, Added missing text, fullname to user account, history recording. (All alarms recorded, view history not implemented.)

alpaslanske:
Full Turkish translation,
Added missing text ("editor.controls-switch")
Added fullname to user account
Added history recording. (All alarms recorded, view history not implemented.)


unocelli:
Hi,
Great job!
I'll try to see if in 'users.fuxap.db' the compatibility can be maintained (for the new column 'fullname')

PR #193, opened 2021-09-03
fixed an issue of incomplete string display when the string starts wi…

span42:
For example, when the tag value is a string of ‘2021-09-03’, only ‘2021’ is displayed


unocelli:
I'm sorry for my delay.
Great job, Thanks. I will merge as soon as possible.

PR #217, opened 2021-11-11
Korean translation

hoonoon:
Add ko.json and modify app-settings.component.ts


unocelli:
Good job. Thanks a lot

PR #369, opened 2022-05-24
Refactor/svg server side lib

AngelFernandezSobrino:
It makes adding new shape groups and SVG easier, both in the front and backend. The frontend method goes out of /lib/svgeditor to be more friendly and uses JSON data files only with SVG data, code required has been refactored to shape.js script.

In the backend, SVG files can be used, svgson is used to create js objects required by SVGeditor. Both methods are loaded to be retro compatible. File 0.json (provisional name) is used to define the shapes group, and two resource endpoints are used to load SVG structure and SVG JSON data. SVG JSON data is processed concurrently to reduce latency, (TODO) a caching method that should be implemented to improve user experience with large SVG databases.

I have also tried to refactor actions, but I have not been able to find how to get SVG content data in GaugeComponent to detect available actions and parse animation code from the SVG file. I have added it inside SVG metadata, but I haven't found how to get it inside the Angular app, sometimes code structure seems a little weird and it may require an effort to improve code structure, maintainability and make it easier to add new features, at least SVG and shapes management.

@unocelli Feel free to remove commits about actions to use on the app, I think that the server-side SVG manager can be useful for the users.


unocelli:
Hi,
I think it is better if you explain me in an online meeting... you can contact me to 4frango@gmail.com

unocelli:
Thanks Ángel, looks a very interesting and promising feature

PR #440, opened 2022-08-04
Input gauge value is now updated every time variable value changes

algismo:
Updating is paused when input has focus (meaning user is editing it). This does not apply when input dialog is in use.
If user leaves input (blur) without changing its value, variable value is written to input in case it has changed while updates were paused.
Also after a key-enter event, input is blurred immediately to make sure that same value is not applied multiple times (if user press the key accidentally multiple times)


unocelli:
Great Job, Thanks a lot.

PR #442, opened 2022-08-05
try parseFloat mqtt payload in order to be able to format number with…

acgiesbrecht:
… significant digits in value output control


unocelli:
Hi, thanks for the PR. However, I think there is a problem with the value is a json object oder a string

PR #456, opened 2022-08-26
"Rotate" action

algismo:
Added new action "Rotate" to the shapes. It sets rotation angle of the shape based on the tag value.


unocelli:
Great job, Thanks a lot

Issue #451, opened 2022-08-12
Input gauge is missing some key functionality


algismo:
Continuing on the input gauge topic (after #431). In my opinion it is missing some key functionality. 
To name:
1. Input validation. Now for example operator can write text in a number field. The result is a "facy" value being sent to the PLC. That is quite crazy!
2. No way to format the shown value. For example for the Text control one can define Units and Format digits, but input gauge has nothing like that. I was wondering why. I think both should have the same functionality available. It is tremendously useful.  
3. When "input dialog" is enabled, current value shall be shown side by side with the new value so that operator has full overview.

The result of all those changes could look like this:

![image](https://user-images.githubusercontent.com/71139387/184444942-d85ab2e9-7c01-4d9a-9eca-96ee08c3aa72.png)
![image](https://user-images.githubusercontent.com/71139387/184445048-9c59e363-bbb4-4e99-a32f-22469330d13d.png)
![image](https://user-images.githubusercontent.com/71139387/184444972-a9f12cb5-c00e-424e-86dd-14bcae6813d1.png)

unocelli:
Great, I like it and it would be nice to add this too.
maybe in the future it will also be useful to add range validation with reference to dynamic values.

unocelli:
Is enable in the last release. I am going to close this as resolved.

PR #463, opened 2022-08-30
Input gauge - missing key functionality

algismo:
Key changes as described in the issue #451.


algismo:
Hi,

I have made a pull request of the implementation I have done for my own needs. Please take a look at let me know what you think.
There are a few places I am not 100% happy:

1. Unit information. Now this is appended to the input as text (like in regular text element), but then unit info must be removed when input field is focused for editing and before applying tag value. My initial idea was to append <div> beside the input element in the editor and keep unit info there, but that was not straightforward since you have not opened your SvgEditor extension file, where the html input is defined.

2. "flex-input" inputType and gauge dlgType. I am reusing flex-input with type "unit", but then in case of input gauge, I also need to show Min/Max fields. I did a hack to find this out in method "isInputMinMax()". Let me know if you will have any better idea how to make this more clean.

3. There is a strange issue when input dialog is enabled and when jumping between input fields with exactly the same value. Then "current input value" disappears since it has not changed (ng model does some nasty thing here my guess). Since it is very corner case, I left it as it is for time being. 

Algirdas

unocelli:
I do some tests with it before merging.

rbonilla31:
Hi, you should try to avoid keyCode in key events, instead use simply key. KeyCode is deprecated. Good job

algismo:
Hi,

correct, I see that. Blindly copied "enter key" check. I will update both.

A side question. I did not manage to find the "source" of the file "ext-bundle.min.js" (in svgeditor dist). Is it intentional? 

Algirdas

algismo:
I did my changes. There is now a conflict due to your recent changes. Shall I incorporate those changes in my branch or you will do manual merge with conflict resolution?

unocelli:
@algismo Hi, I have to look and maybe do some modifications, I would appreciate if you could make the PR to the branch issue-451 that I just created.

algismo:
Hi. I am not sure what to do :). Some days ago you did a merge into my branch "@unocelli
[Merge branch 'master' into issue-451](https://github.com/frangoteam/FUXA/pull/463/commits/f02f3fc9f2604cdabbeaf2c2e51633666642c2cd)". Now there are no conflicts so PR can be merged into master. Are there any other issues?

PR #468, opened 2022-09-02
allow scripts, forms and modal inside iFrame

mauroalexandre:
today iFrame can't submit forms and execute scripts, this is important feature when you use another app inside yours


unocelli:
Thanks :+1:

PR #486, opened 2022-09-14
Add "mode execute (client/server)" capability in scripts.

mauroalexandre:
Sometimes we need to use script for frontend operations, like change color, values, etc by selector.

This feature add "Mode" characteristic for scripts where user can choose where the script is evaluated.

**Server** _(default)_ - Keep the same behavior, call /api/runscript and execute the source in node backend.

**Client** - Execute script in client browser.

I'm using this option to solve alot of problems in my HMI. I dont know if this feature are welcome to te project, but here my loving PR.

Thanks for yout project.


unocelli:
Great job, this feature has many potentials, I love it.
Thanks a lot.

PR #496, opened 2022-09-17
add settings for auto resize view

mauroalexandre:
create a setting to fuxa auto resize view when window size is changed


unocelli:
Thank you very much for the PR

client/src/app/_helpers/utils.ts
```diff
@@ -296,6 +296,19 @@ export class Utils {
...
+    static resizeView = (selector) => {
+        document.querySelectorAll(selector).forEach((scaled: any) => {
+            let parent = scaled.parentNode,
+            ratio = (parent.offsetWidth / scaled.offsetWidth),
+            padding = scaled.offsetHeight * ratio;
+      
+          scaled.style.transform = 'scale(' + ratio + ')';
+          scaled.style.transformOrigin = 'top left';
+      
+          parent.style.paddingTop = padding;
```
unocelli, line 309: Hi, why you set the paddingTop?

client/src/app/_helpers/utils.ts
```diff
@@ -296,6 +296,19 @@ export class Utils {
...
         }, {});
     }
+
+    static resizeView = (selector) => {
+        document.querySelectorAll(selector).forEach((scaled: any) => {
+            let parent = scaled.parentNode,
+            ratio = (parent.offsetWidth / scaled.offsetWidth),
+            padding = scaled.offsetHeight * ratio;
+      
+          scaled.style.transform = 'scale(' + ratio + ')';
```
unocelli, line 306: @mauroalexandre suggestion:
```
            let parent = scaled.parentNode,
            ratioWidth = (parent.offsetWidth / scaled.offsetWidth),
            ratioHeight = (parent.offsetHeight / scaled.offsetHeight);
      
            scaled.style.transform = 'scale(' + Math.min(ratioWidth, ratioHeight) + ')';
```

client/src/app/_helpers/utils.ts
mauroalexandre: Sorry, I need to remember, but, I think your solution is better.

client/src/app/_helpers/utils.ts
mauroalexandre: Nice!

PR #500, opened 2022-09-19
Feature/script mode client schedule support

mauroalexandre:
Now the schedule functions works in client mode too.


unocelli:
Super work, Thanks a lot

Issue #487, opened 2022-09-15
Select value gauge. Different fill and stroke color for each item.


algismo:
"Select value" gauge is a very useful control both as an input and as a status indicator when used in "Read Only" mode.
The only problem with it is that all choices look the same (font color, background). This becomes a significant issue if control is used to show state.

I have extended it to allow user to choose font and background color (optional) for each item:
![image](https://user-images.githubusercontent.com/71139387/190321649-f349d13e-41ab-4e33-bd34-a00c6564d57b.png)

Will make a merge request.

unocelli:
I am going to close this as resolved. let me know if you have any issues.

PR #530, opened 2022-10-09
Added background and text color choice for each option in select gauge.

algismo:
See #487.


unocelli:
Hi, Thanks for the PR, Good job

Issue #535, opened 2022-10-12
Multiple tag selection


AlexandruAndrei3:
Hi,

When i have a large number of tags in order to add them i have to click to select  each one of them (eg: If  have 1000 tags i have to click 1000 times in order to add them). I think it can be useful to add an Select All option sou users don't have to select each tag one by one.
![Capture2](https://user-images.githubusercontent.com/94168341/195308646-e0d8472d-d069-4cfd-b696-152de38205b6.PNG)

mauroalexandre:
Hi! Thanks for the sugestion.
I developed this possibility.

Please check #536

unocelli:
I am going to close this as resolved. let me know if you have any issues.

PR #536, opened 2022-10-12
 Add option to multi select tags (#535)

mauroalexandre:
Now users can use root node for check/uncheck all childs.

#535


haitan981:
> Now users can use root node for check/uncheck all childs.
> 
> #535

After editing the files according to your instructions. Do I need to reinstall Fuxa ? Because I saved and pressed F5 to reload the page but no change in the web interface.

mauroalexandre:
@haitan981 when you use source from github, your frontend use dist from the last build.

For you see changes in your project before wait the release of the next version, you need to build your client.

cd client/
npm install
npm build

unocelli:
Hi, Thanks a lot, Good job :D

PR #548, opened 2022-10-16
Fix translation of editor alignment

Bykiev:
Fix translation of editor alignment


unocelli:
Hi, Thanks for the PR, Good job

PR #643, opened 2022-12-11
first french translate

lbeyaert65:
add french to fuxa


unocelli:
Good job, Merci beaucoup :D

PR #770, opened 2023-04-01
Add allow-same-origin flag to solve CORS problems using different ports

mauroalexandre:
add allow-same-origin in sandbox attribute from iFrame to allow open page in the same domain with different ports


unocelli:
Thanks a lot

PR #780, opened 2023-04-11
Added support to Modbus RTUOverTCP.

rvbatista:


unocelli:
Hi, Thanks a lot for the PR, looks good.
I wonder if it would not be better to use an extra attribute in the options to identify the connection type for modbus, so we can have RTU with options or TCP with options.
ModbusRTUOverTCP and ModbusTCP has the same settings! you could add a selector with TCP (default) and RTUOverTCP. what do you think about?

unocelli:
let me know, will your modification for RTU over TCP with connectTelnet work?

rvbatista:
I was not able to test with real devices, but I was able to see that FUXA was sending Modbus RTU over TCP connection with a telnet server.

PR #869, opened 2023-07-19
Update README.md

OptifySudarshanPatil:
Added extra instruction to run git clone installation. otherwise server will not start and error will prompt due to missing library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,9 @@ Or [Download the latest release](https://github.com/frangoteam/FUXA/releases) an
 ```
 cd ./server
 npm install
+cd ..
+cd ./server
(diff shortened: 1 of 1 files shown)


unocelli:
Hi, sorry for the delay.
I don't understand, the error must be for another reason.

OptifySudarshanPatil:
Hi,
i tried and verified this with 2 different windows system.
if not server nevre run.

unocelli:
this is correct, in fact in the description once I enter in the server folder I first do install and then start.
in your PR you enter into the server folder, do the install then exit from server, then enter again into server folder and then do the start

PR #889, opened 2023-08-08
fix undefined name in daqstore for influx2

mauroalexandre:


server/runtime/storage/influxdb/index.js
```diff
@@ -112,7 +112,7 @@ function Influx(_settings, _log) {
             } else {
                 const point = new Point(tag.id)
                     .tag('id', tag.id)
-                    .tag('name', tag.name)
+                    .tag('name', tag.tagref.name)
```
unocelli, line 115: Hi thanks for the PR, I'm not sure if all devices set the tagref, what do you think about my suggestion
`.tag('name', tag.name || tag.tagref.name)`

mauroalexandre: perfect

unocelli:
Thanks a lot :wink:

Issue #882, opened 2023-08-04
German translation


EtherGraf:
I intend to create a german translation file. 

Is there any support in doing so, or does it possibly already exist in the pro version?

unocelli:
Hi, the German translation is missing, would be great...you have to add the fuxa\client\src\assets\i18n\de.json file,  add `{ text: 'dlg.app-language-de', value: 'de' }` in [languageType](https://github.com/frangoteam/FUXA/blob/master/client/src/app/editor/app-settings/app-settings.component.ts) and to be able to test it you have to be able to compile the frontend in angular.

EtherGraf:
Created a pull request #894.

unocelli:
Thanks again

PR #894, opened 2023-08-14
Added a german translation file de.json

EtherGraf:
As intended in #882 here is the text resource file with the german translation.

The angular build is not yet run in this pull request, because I have it not installed on this system.


EtherGraf:
I am unsure if the key 'lg.app-language-de' is needed in all other languages and how to translate it.

unocelli:
Sieht sehr gut aus, vielen Dank :rocket:

PR #897, opened 2023-08-22
fix: image resources not found in dev mode

k1nz:
Hi, i found the custom image resources was failed to load in dev mode. I think that the image path should be prefixed with the server address.

b077c0c8088cb7b57cb1dbaa27532eb

fe171f65399e85c16ce26a8f8bd396b


zsdycs:
Very good! I also discovered this issue and used this code to fix it.

image

unocelli:
Hi, great job, Thanks a lot.

PR #901, opened 2023-08-24
fix: opcua connect error: End point must exist

zsdycs:
## The specific problem to be solved can be seen in the screenshot below:

微信截图_20230824141943

微信截图_20230824141644

Only after modification can OPCUA be connected.

Specific questions can be referred to [here](https://github.com/node-opcua/node-opcua/issues/63).


unocelli:
Hi, great job, Thanks

PR #918, opened 2023-09-27
Shenqiax

shenqiGitHub:
测试


concept10:
Out of curiosity without reviewing the changeset, what is the new s7 connection type for?  (since s7 device drivers exist). Thanks in advance.

wmcsl:
In https://github.com/frangoteam/FUXA/pull/918/commits/b6651c5944ab93627540144d8b19d6dc23bc82c9#diff-277c5d674bf85848e67ebab280071c5eed6203f48607586d61f704463eafca8fR63 it appears that administrative functionality is solely handled by checking if the username is "admin". Is that the intended authorization framework for this software?

unocelli:
If look good into the code you see that this validation is only to block in the frontend to remove the admin user. what is wrong in this?

Issue #995, opened 2023-12-03
[BUG] Confguration of InfluxDB address 


robbudge:
Testing  on Version 1.1.17-1370

When configuring Daq Connection Influx 2+,
Failure to Specify protocol in Server URL will break the runtime after a  'Restart' Command. Leaving no accesses to the system.
Daq Server URL 192.168.10.13:8086 would result in failure.
/server/_appdata/mysettings.json has to be manually edited too include http:// or https:// 
http://192.168.10.13:8086

Perhaps a validation check could be included to ensure https:// or http:// is included in the Daq Servers URL.

robbudge:
I was modifying the catalog setting on the V1.17 release and omitted the http:// prefix for my InfluxDB server. 

Obviously I got lots of errors. 
thinking it was a change error I restarted the Fuxa Process and it failed to launch and exited on startup. 
knowing what I had just changed, I modified the base file with the correct http:// prefix. 
restarted and all was good.

mauroalexandre:
#1062

unocelli:
This issue should be fixed in the latest version. let me know if you have any issues.

PR #1062, opened 2024-01-23
Bypass daq connection if failure (#995)

mauroalexandre:
I'm not sure if it's the optimal solution, but as it stands, if the DAQ connection fails, users are unable to access Fuxa to rectify the incorrect address. Implementing a bypass would grant users the ability to do so.

Fix for #995


unocelli:
Good Job!, Thanks a lot

Issue #1033, opened 2023-12-26
[FEATURE] Add the 'name' attribute to the element when rendering the view as an SVG.


mauroalexandre:
To make it easier for a frontend script to find an element through the selector, it would be fantastic if the name of the element was in some SVG attribute created when the view is rendered.

mauroalexandre:
@unocelli hello, does this pull request make sense to you? #1103

If so, it would be interesting to be able to apply the same attribute to elements such as Rect, Text, and especially to svg-ext-shapes-image.

With this, it would be possible to switch between images, texts, etc., through the execution of scripts using selectors.

unocelli:
It is available in the latest version, Thanks!

PR #1103, opened 2024-03-06
Feature: Add Tagging Capability for Element Selection via Selectors (#1033)

mauroalexandre:
Introduce a `data-gauge-name` attribute for elements to simplify manipulation through selectors in scripts.


unocelli:
Hi, good job. I will add to the other control too.

PR #1147, opened 2024-04-06
ODBC Driver Install 

MatthewReed303:
I have added an odbc folder containing an install script and odbcinst.ini with odbc driver configs. I have also updated the docker file to node: 16-bullseye so we have better support for arm64 on RPI. I have also added the build of node-odbc as there are no prebuilt packages for arm. The build does not work on node: 18 and needs to be investigated.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:18
+FROM node:16-bullseye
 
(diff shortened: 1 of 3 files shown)


unocelli:
Look really good

PR #1149, opened 2024-04-09
Enhanced Device Polling Rate

henjoe:
On the device PollingRate, I noticed that if we set the rate into 5secs, and the Fuxa server start at 08:02, then the next polling rate will be 08:07, 08:12 , 08:17 and so on and so forth.. 

This request ensures that the polling rate will always be modulos to the pollingRate settings in the device. 
If you set the rate to be 5secs, and the Fuxa server start at 08:02, then the next polling will be at 08:05, 08:10, 08:15 and so on and so forth... 
Setting the pollingRate into 10 secs, then the polling will be modulos of 10 -> 08:10, 08:20, 08:30 and so on and so forth.

PS: Really new to Github, this is my first time to pull request, Apologies if I did something wrong or mistakes. 😄

### DIFF
--- a/server/runtime/devices/device.js
+++ b/server/runtime/devices/device.js
@@ -178,10 +178,21 @@ function Device(data, runtime) {
             comm.init(MODBUSclient.ModbusTypes.TCP);
         }
         return comm.connect().then(function () {
-            devicePolling = setInterval(function () {
+            devicePolling = bindClockTick(function(ms) {
                 self.polling();
-            }, pollingInterval);
+                // console.log(data.name + " - > pollingInterval: " + pollingInterval + " " + new Date().toLocaleString())
+            }) 
         });
+    }
+
+    function bindClockTick(callback) {
+        var SECOND = pollingInterval;
+        function tick() {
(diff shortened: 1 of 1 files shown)


unocelli:
Hi, Thanks for the PR.
I understand the issue, but I wonder if with this kind of approach we don't risk resource or performance problems in continuing to create timers?

henjoe:
Hi @unocelli , to be honest not really sure and haven't tested yet the performance about this. You might have good point on that. 
How about this code here? I think this much similar to the existing except that we added condition to check if the current seconds in modulos to pollingInterval:

`devicePolling = setInterval(function() {
                var now = new Date()
                if (now.getSeconds() % (pollingInterval/1000) === 0){
                    console.log(data.name + " " + new Date().toLocaleString() + "-->> " + (now.getSeconds() % (pollingInterval/1000) === 0))
                    self.polling();
                }              
            },1000)`

unocelli:
Hi, I suggest to make it configurable in _appdata/settings.js

PR #1170, opened 2024-04-26
Update Dockerfile

MatthewReed303:
remove odbc branch tags as it has now been merged into master


unocelli:
We will add odbc in the next version with update to nodejs 18

MatthewReed303:
@unocelli no problem, there is a new PR for the dockerfile from 2 weeks ago where the full build is self contained, I recommend pulling that one.

Issue #1197, opened 2024-05-18
[FEATURE] tag option - scale by script 


greg9504:
**Describe the feature**
Tag options for scaling values are currently limited to 3 methods (linear transform, ms to date, ms to time).  There are situations where additional transforms must be done to the value before being sent to the client.  

Allowing a script to transform the value would allow for support of more hardware without changes to device plugins.  

As an example SEW MDX61B* Variable Frequency Drives MODBUS requires parameter values to be encoded in an 8 byte sequence:
![image](https://github.com/frangoteam/FUXA/assets/1952281/a0526b88-5a0f-40d3-a6af-844bf44cc58e)

While it is possible to assign a script to an object (button etc) event to transform the value, script events are not available for all object types (sliders).   

**Describe the solution you'd like**
In the tag options dialog, add a scale script selection.  The user can then select the script to apply the value transform.
![image](https://github.com/frangoteam/FUXA/assets/1952281/5e8ac9e9-caa9-423d-b90c-7652cc6c8e29)

The script must adhere to the simple format of one input variable, and it must return one value, and be configured as SERVER.  The format of the output is device plugin dependent.  In the case of the MODBUS device, the output would be Node Buffer.

![image](https://github.com/frangoteam/FUXA/assets/1952281/9ae59734-37af-4d87-b70b-643522341130)

![image](https://github.com/frangoteam/FUXA/assets/1952281/83586e5d-e762-4ff6-beb0-50367d1b85e1)

```javascript
async function buildSewRampSpeedBuf (value) {
    //movilink parameter channel format:
    // [0x32] management byte = 0x32  
    // [0x00] parameter sub index byte (usually 0x00)
    // [0x2116] parameter index, two bytes 
    // [0x00000000] 4 byte parameter value
    const valBuf = Buffer.alloc(8, 0);
    valBuf.writeUInt8(0x32);
    valBuf.writeUInt16BE(8470, 2);
    valBuf.writeUInt32BE(value, 4);
    console.log(`buildSewRampSpeedBuf returning buffer: ${valBuf.toString('hex')}`);
    return valBuf;
}
```

**Additional context**
The scale script would be run after any other configured scaling for the tag and is independent of the scale options.

example for MODBUS device in setValue
/server/runtime/devices/modbus/index.js
```javascript
var memaddr = data.tags[sigid].memaddress;
            var offset = parseInt(data.tags[sigid].address) - 1;   // because settings address from 1 to 65536 but communication start from 0
[...]

PR #1208, opened 2024-05-30
Add option for read/write scripts for tags

greg9504:
As discussed in this issue:
https://github.com/frangoteam/FUXA/issues/1197

I've tested with 
bacnet (see below for additional notes)
http client
modbus
mqtt
opcua

General script calling
In msm.js lines 50 and 59 you had:
var paramsValue = _script.parameters.map(p => p.value || p);

when p.value === 0, this would pass p.  I assume that is not wanted and 0 should be passed to the script.  I changed it to:
var paramsValue = _script.parameters.map(p => _isNullOrUndefined(p.value) ? p : p.value);

perhaps it should only check for undefined?

Bacnet:

While testing the various clients I noticed that there might be a problem with the way the Bacnet polling is done.  
The polling method in Bacnet does:
await client.readPropertyMultiple(...)
but the node-bacnet client is not async.  

Tag scripts use the existing methods for calling scripts, which makes all scripts async.  So I've changed all the code to handle this.  The problem with how the bacnet client works is that the polling routine can be called multiple times while still waiting for the result from a previous call.

Example:  User configures 3 second polling, but response takes longer than 3 seconds.
Time
0s) polling is called ->  client.readPropertyMultiple(..., callback) returns immediately --> polling function returns 
the callback passed to client.readPropertyMultiple has not returned.
3s) polling is called ->  client.readPropertyMultiple(..., callback) returns immediately --> polling function returns 
the callback passed to client.readPropertyMultiple has not returned.
the callback for the first call to client.readPropertyMultiple returns...
6s) polling is called ->  client.readPropertyMultiple(..., callback) returns immediately --> polling function returns 
the callback passed to client.readPropertyMultiple has not returned.
the callback for the second call to client.readPropertyMultiple returns...

Basically the _checkWorking is not doing what it is supposed to. 

To solve this I wrapped the call to client.readPropertyMultiple in a promise.  This results in:

0s) polling is called ->  client.readPropertyMultiple(..., callback) returns immediately --> waits for response...
the callback passed to client.readPropertyMultiple has not returned.
3s) polling is called, _checkWorking fails
'BacnetTest' working overload! 0
polling exits
the callback for the first call to client.readPropertyMultiple returns...
the first polling call returns and clears the working flag
6s) polling is called ->  client.readPropertyMultiple(..., callback) returns immediately --> waits for response...

Other changes in the bacnet client:
changed the 
value.values.forEach(data => {
to
value.values.forEach(bacData => {
as data was too confusing with the data member
removed the errors[] array, wasn't used.
With the bacnet device simulator I was using, errors are indicated by  
bacData.values[0].value[0].type === bacnet.enum.ApplicationTag.ERROR
where as the current  check is done for
bacData.values[0].value.type === bacnet.enum.ApplicationTag.ERROR

I've changed the code to look for both, but perhaps this needs more testing if only one is correct. I was using the Bac simulator from SCADAEngine.com

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -29,7 +29,7 @@
       "type": "chrome",
       "request": "launch",
       "url": "http://localhost:4200/#",
-      "webRoot": "${workspaceFolder}/client/src"
+      "webRoot": "${workspaceFolder}/client/"
     },
     {
       "type": "node",
--- a/client/src/app/_helpers/utils.ts
+++ b/client/src/app/_helpers/utils.ts
@@ -1,5 +1,4 @@
 import { Injectable, Pipe, PipeTransform } from '@angular/core';
-import { AbstractControl, ValidationErrors } from '@angular/forms';
 import { DomSanitizer } from '@angular/platform-browser';
 
 declare const numeral: any;
@@ -557,4 +556,3 @@ export class EscapeHtmlPipe implements PipeTransform {
         return this.sanitizer.bypassSecurityTrustHtml(content);
     }
 }
-
--- a/client/src/app/_helpers/utils.ts
+++ b/client/src/app/_helpers/utils.ts
@@ -1,4 +1,5 @@
 import { Injectable, Pipe, PipeTransform } from '@angular/core';
+import { AbstractControl, ValidationErrors } from '@angular/forms';
 import { DomSanitizer } from '@angular/platform-browser';
 
 declare const numeral: any;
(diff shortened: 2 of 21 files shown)


unocelli:
Hi @greg9504, I'm reviewing your PR.
Do you see any problems if I move the attributes to TagScale and add the ScriptMode to TagScaleModeType?
That way I can only choose one type of scaleMode:

image

 so I can also disable the functionality

greg9504:
Hello Umberto,

I wrote the code to allow both options to be used on a tag.  On send (device-utils.js 
tagRawCalculator) the value has any configured scale method applied, then the script called.  On read (device-utils.js tagValueCompose) the script is called first then the scale function.  I figured that could be useful.  If the device requires some extra formatting it can be done by the script, but scale can still be handled by FUXA.  

Example case: For end users who may not know javascript that well, they can get help with code to apply/remove the extra formatting but not have to worry about coding the scale function as well, which would be different for each tag.  

The tag script read/write drop downs have an empty row so that it can be set to none.

unocelli:
Hi Greg, all right, excellent work! ❤️ 
Just making a few styling changes in frontend...thank you very much for doing some code reviews of my code.

PR #1229, opened 2024-06-10
Fix alarm counter in header to display only active alarms

mauroalexandre:
Currently, the alarm counter in the header sums all alarms regardless of their status, including those with the status **AlarmStatusEnum.VOID**, resulting in an _incorrect_ number. 

This pull request modifies the counter to display only the **ACTIVE** alarms. This behavior is standard in other tools, but we can improve it in the future to make this behavior configurable.

### DIFF
--- a/server/runtime/alarms/index.js
+++ b/server/runtime/alarms/index.js
@@ -72,7 +72,9 @@ function AlarmsManager(_runtime) {
                 var result = { highhigh: 0, high: 0, low: 0, info: 0, actions: [] };
                 if (alrs) {
                     Object.values(alrs).forEach(alr => {
-                        result[alr.type]++;
+                        if (alr.status == AlarmStatusEnum.ON) {
+                            result[alr.type]++;
+                        }
                         if (alr.type === AlarmsTypes.ACTION && !alr.offtime) {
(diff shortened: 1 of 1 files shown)


unocelli:
Your modification only counts alarms that are in status ON, for me OFF and ACK status must also be counted, therefore all those in the alarms table, alarms are then removed from the table automatically according to the mode. Maybe the issue you are having is a different one and there are 'ghost' alarms left in the table (not removed for another issue).

LordNetro:
It happens to me whenever I press the ACK all only, any thoughts?

MatthewReed303:
I can confirm I have issues with the alarm count also, if you ACK one alarm at a time it will be correct, if you ACK all it stuffs up the count. This has be a complaint from multiple customers. The alarm count indicator should only display active alarms and that's it

unocelli:
Thanks for the report, the issue should be fixed

unocelli:
can you please confirm that the problem has been solved

mauroalexandre:
> Hi, can you please confirm that the problem has been solved

Solved, thanks. In sometimes status = '', your codeguard fixed this.

Thanks alot.

PR #1242, opened 2024-06-21
Update install_odbc_drivers.sh

MatthewReed303:
remove exit and continue installation on armhf ( MySQL ODBC driver does not support 32bit )


unocelli:
Good job, Thanks a lot!

Issue #868, opened 2023-07-19
Switch button changes its size while being used.


Bundyra:
Hello, today i found the issue with switch button. If you change the button's default size of 50x28 to any other size, then pressing the button causes it to change its size/position (not sure which one). Example: i changed buttons size to 30x18 and this what happens after pressing it:
![image](https://github.com/frangoteam/FUXA/assets/79423733/2e2cb4b0-5f16-4a83-9c32-4faf8fb1fc00)

unocelli:
Hi, We can't reproduce it...can you send me an example project?

Bundyra:
Sure, I sent an email with an attachment.

unocelli:
Hi, Thanks, I can reproduce it and we will fix it as soon as possible.

WeiFengHong:
Hi, is there any progress on this issue? There is still a size change problem when switching status.

PR #1288, opened 2024-07-22
Fix: Incorrect Display Height of Toggle Switch

WeiFengHong:
#868 
This fix addresses an issue where the height of the toggle switch was not displaying correctly. The problem was resolved by setting the vertical-align style property to middle for the input element.


unocelli:
Hi, thanks for your try but your fix is not working, tested with 8px height in Chrome

PR #1312, opened 2024-08-07
Feat/modebus tcp reuse

rikugun:
sometimes, modbus slave not support multi connect at same time. so we need reuse socket connection.

image

diff with open socket reuse

image


unocelli:
Hi, Good job, sorry for the delay, I'm reviewing your PR.

unocelli:
@rikugun sorry for the delay, Good job, Thanks a lot

PR #1325, opened 2024-08-20
Round the number before converting it to a BigInt for ModBus

Abestanis:
Hi, thank you for this awesome project, this is an incredible powerful software!

I ran into a problem while trying to use a `Divisor` with an `Int64` ModBus value. Reading the value works fine, but writing was throwing the following error:
```
RangeError: The number 9999.99998 cannot be converted to a BigInt because it is not an integer
```

`BigInt` only accepts integers, so in my opinion the correct behavior should be to round the value to the closest integer when `toCast` is true, which is only the case for integer types so this won't affect float types. I tested it with these changes and it fixed the problem for me.

There is also some whitespace changes from my editor, let me know if you want me to remove them.

### DIFF
--- a/server/runtime/devices/modbus/datatypes.js
+++ b/server/runtime/devices/modbus/datatypes.js
@@ -2,14 +2,14 @@
 function _gen(bytes, bFn, WordLen, swapType, toCast = false) {
     return {
         bytes,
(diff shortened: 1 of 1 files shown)


unocelli:
Good job, Thanks a lot

PR #1333, opened 2024-08-25
Update odbc to v2.4.9 and node to v18

MatthewReed303:
odbc v4.2.9 fixes install bugs, but requires node v18 have also removed the odbc build and install in the dockerfile as it's no longer needed


unocelli:
Thanks a lot

PR #1347, opened 2024-09-02
Added 350ms Polling Interval Option for PLC

mauroalexandre:
This pull request introduces the option to set the PLC polling interval to 350ms. Although this might seem like a minor change, in certain scenarios—like the one I'm encountering—this adjustment in polling time is critical and could benefit other users of the project.

Thank you for considering this change, and I'm open to discussing it further.

Thank you!

### DIFF
--- a/client/src/app/device/device-property/device-property.component.ts
+++ b/client/src/app/device/device-property/device-property.component.ts
@@ -29,6 +29,7 @@ export class DevicePropertyComponent implements OnInit, OnDestroy {
 	showPassword: boolean;
 
 	pollingPlcType = [{text: '200 ms', value: 200},
+					  {text: '350 ms', value: 350},					  
 					  {text: '500 ms', value: 500},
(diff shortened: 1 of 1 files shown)


PedroHDDomingos:
Important improvement for small PLCs where we normally cannot work with 200ms.

MatthewReed303:
I feel we should start from 100ms with 50ms increments or have ability to enter in custom polling number. High performance systems can do 100ms no problem and great for real time motion control applications where trending updates and PID tuning is important on high speed systems. What are your thoughts on this @unocelli

PedroHDDomingos:
Excellent @MatthewReed303, I believe that 50 to 50ms would be enough, today there is a jump from 200 to 500ms, this makes fine adjustment very difficult.

unocelli:
I am positively for adding 50, 100, 150, 350

unocelli:
@mauroalexandre Thanks, I will merge this PR and add later the other values.

PR #1356, opened 2024-09-04
Add Option for Auto-Selecting Text in Input Fields on Click

mauroalexandre:
In most popular SCADA software systems, the standard behavior when clicking on input elements is to automatically select all the text. This facilitates easy replacement of the text. Following this common usability practice, this pull request introduces an option that allows users to enable this behavior for input elements within our application. A demonstrative video is included below to show how this new feature works.

https://github.com/user-attachments/assets/55cdd344-9e9e-449c-ac0b-ed0ceca3c11e

### DIFF
--- a/client/src/app/_models/hmi.ts
+++ b/client/src/app/_models/hmi.ts
@@ -213,6 +213,7 @@ export interface InputOptionsProperty {
     timeformat?: InputTimeFormatType;
     convertion?: InputConvertionType;
     updatedEsc?: boolean;
+    selectOnClick?: boolean;
 }
 
 export enum InputOptionType {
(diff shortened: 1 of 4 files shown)


MatthewReed303:
@mauroalexandre awesome job! I was discussing this feature with @unocelli, can you test if it works using the virtual keyboards also?

mauroalexandre:
Hey @MatthewReed303 , thanks.

Not tested in virtual keyboard yet.

unocelli:
@mauroalexandre Good job! Thanks a lot.

MatthewReed303:
@mauroalexandre one thing I have noticed when the screen first loads with a value already in the input, it does not work, but after clicking in, then out and back in it then works correctly. Whether this is because of the updates being enabled or just the value there and not blank? 

I can confirm it also works with the virtual keyboard fine.

Issue #1175, opened 2024-05-02
[BUG] ModbusTCP 16bit register write - bad function code


AllmarkPL:
Hello,
I'd run into an issue where I can't write values into single 16bit register of modbus device. I have created a single tag with INT16 datatype and proper device addressing, but when I tried to write data into it nothing happended. Quick look into modbus logs has shown that function 16 (multiple registers) is used instead of 06 (single registers) and communication is interrupted due to "bad lenght 8 (expected more than 8)". When I switched Fuxa tag datatype to INT32 all started working, but I can't leave it that way, because device I'm using have 6 consecutive 16bit registers which I got to use... Is there a way to properly send single 16bit register by modbus TCP in Fuxa? I don't think it's issue with the device itself, because I can write data into it using simulator or HMI.

**To Reproduce**
1. Create a modbus TCP IP device with single INT16 tag
2. Try to write any value either due to script action or by input.

**Expected behavior**
I think it may be a bug in modbus tcp plugin.

**Environment**
 - FUXA  Version 1.1.19-1524
- latest docker container
 - Platform/OS: Server: Ubuntu; Client: WIN11
 - Browser: Chrome 123.0.6312.86

rvbatista:
Hello, thanks for your issue.
Are you able to test both conditions with Wireshark capturing the packages?

AllmarkPL:
Hello,
as for now I can provide packet data for writing into modbus simulator software, I will have access to physical device on monday.
As You can see on packets below, the function code used for writing is in fact 16, and is lacking second byte in transmission which causes error. First screen is from writing by script, second is from writing by input box. Third is after changing datatype to INT32.
I think, that it was working properly in one of older versions, but I got to verify that.
Writing to INT16 by input box:
![image](https://github.com/frangoteam/FUXA/assets/168782821/df6bdcc0-8b65-42b9-90cf-ae12a827f7ce)
Writing to INT16 by script:
![image](https://github.com/frangoteam/FUXA/assets/168782821/b4b22296-20c7-46a5-92c6-775b321b68ec)
Device dataset:
![image](https://github.com/frangoteam/FUXA/assets/168782821/89c9f124-b696-4f88-9ad5-ccd0d3829b47)
Modbus dump:
![image](https://github.com/frangoteam/FUXA/assets/168782821/cf6a432e-e37d-4d1f-8fdb-df4aa650628a)
Writing to INT32 by the same script:
[...]

PR #1364, opened 2024-09-09
Update index.js Issue #1175

rvbatista:
Issue #1175
should be something like this, but I was not able to test.


rvbatista:
I'm not sure, you see depends on the variable type also, INT should use WriteRegister, and any double REAL, DINT, etc.. should always use WriteRegisters.

As fuxa was only using WriteRegisters, I think that it's interesting to have a user accessible setting to control this behavior.

Fragment feature is already used by Modbus?

unocelli:
The Problem is that value is a bytes array, 
`var _writeMemory = function (memoryAddress, start, value) {`

Int16, Uint16 = 2 bytes
Int32, UInt32 = 4 bytes
Bool = byte

If we define that 1-2 with WriteRegister and 4-8 with WriteRegisters will work for every device? is modbus standard?

rvbatista:
For me both should work, but by specification there is the function to write single register (WriteRegister) and Write Multiple Registers (WriteRegisters) both are 16 bits based for each register.

unocelli:
Thanks a lot.

PR #1369, opened 2024-09-09
Fix for Concurrent Blink Actions Across Different Ranges

mauroalexandre:
This pull request addresses an issue where multiple blink actions fail to execute properly when their conditions are met across various ranges. The root cause was identified as the non-restarting of the blink timer when updates occur within the current range, which inadvertently prevents other blink actions from executing.

## Changes Made

- Modified the logic in `gauge-base.component.ts` to ensure that the blink timer is restarted whenever necessary, allowing for concurrent blink actions across different ranges without interference.
- Added checks to ensure that the timer restart logic does not impact performance or cause unexpected behavior in other components.

## Impact

These changes allow all blink actions to function independently and as expected, regardless of the range conditions. This fix enhances the functionality and reliability of the gauge components by ensuring that visual cues are accurately represented in real-time.

## Testing

- Comprehensive testing has been conducted to verify that the changes resolve the issue without introducing new bugs.
- Tests included scenarios where multiple blink actions are triggered across overlapping and distinct ranges.

Please review the changes and provide feedback or approve the pull request for merging.

### DIFF
--- a/client/src/app/gauges/gauge-base/gauge-base.component.ts
+++ b/client/src/app/gauges/gauge-base/gauge-base.component.ts
@@ -108,7 +108,8 @@ export class GaugeBaseComponent {
         }
         gaugeStatus.actionRef.type = act.type;
         if (toEnable) {
-            if (gaugeStatus.actionRef.timer) {
+            if (gaugeStatus.actionRef.timer && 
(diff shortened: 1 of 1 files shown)


unocelli:
Good job, Thanks a lot

henjoe:
Can you confirm that this works on this setup ?
image

Because when I tried, it only works on the 2nd action blink, the first blink action doesn't fire.

PR #1382, opened 2024-09-14
bug fix for tdenginer DAQ: add tagid in SQL query as filter condition

jay-jjwu:
Hi, 
when i test trending of multi-curves, retriving data from tdengine database, i found some data does not belong to this trending shows up.
i found the tagid is missing as the filter condition in SQL query in tdengine, that's why multi-data are show trending in chaos.
i think it might be a bug, and i add into the SQL query.
Please check the PR.
image
image


unocelli:
Hi, Thanks. Please check the PR, you have other features or change that I can't merge. I need a PR only with the fix.

jay-jjwu:
sorry, i just found i submit PR based on wrong branch, some of the features i'm still in testing. i will take notice of this in next PR.
Thanks!

PR #1404, opened 2024-09-22
bugfix: for tdengine database, in sql query, the tagid is missed as o…

jay-jjwu:
hello, here's the PR with only this fix of tdengine bug. Please check.
Thanks!


unocelli:
Good job, Thanks a lot!

PR #1412, opened 2024-09-27
Feat/modbus tcp mutex

### LINKED ISSUE #1343: [BUG]_readMemery Error When Enabling Multiple Devices with 150 Tags Each
**Describe the bug**
I have configured two devices, which together have approximately 150 tags. Each device is connected to the server using ModbusTCP. Additionally, the two devices are connected to each other using RS485 through a serial server, which is then connected to the company’s internal network.
image
image

When I enable either device individually, Fuxa works normally without any issues. However, when I enable both devices simultaneously, the log persistently reports errors.

image

I have already tried the solutions mentioned in issues #154 and #191, but unfortunately, they didn't resolve the problem. The errors still occur when both devices are enabled at the same time.

Additionally, I attempted to enable the fragmented flag, but this caused another issue: the application was unable to read all tag values properly.
image

Could you please provide further assistance or suggest any additional steps I can take to resolve this issue?

Thank you!

**Environment**
 - FUXA version: 1.2.0
 - Node.js version: 
 - npm version:
 - Platform/OS: docker
 - Browser:chrome 128.0.6613.114

rikugun:
fix #1343

For modbus-tcp,  if open socket reuse ,  add option to read/write serial.

image


unocelli:
Thanks, I do a review and then merge it

PR #1413, opened 2024-09-27
fix:  setTagValue  always return null

rikugun:
device.setValue has changed to async func , then setTagValue always return null;

QQ_1727428250038


unocelli:
Good job, Thanks a lot!

PR #1480, opened 2024-11-05
fix: webAPI http request Error if value is null

ChimingLiu:
when request value is null, `items[key].type` will go wrong

server\runtime\devices\httprequest\index.js

fuxa-err.log
`2024-11-05T14:10:58.479Z [error] 'https://xxx/test/1' _readRequest error! TypeError: Cannot read properties of null (reading 'type')`


unocelli:
Good work. Thanks a lot!

PR #1520, opened 2024-12-04
Dev chart zone

Kiruthick-Roxx:
Option to configure zones for both fill and stroke with individual zone arrays. Also added a function to dynamically change the tooltip marker color based on the tagvalue at that time period.


unocelli:
Hi, good work but I have a question, why do you separate stroke and fill into two lists? fill is also associated with a line like stroke, I would simply add the fill attribute to ChartLineZone

Kiruthick-Roxx:
Hi,
   The problem is, when the color attribute is applied to both stroke and fill, while plotting, the graph looks weird and it becomes difficult to differentiate when 2 or more tags are added for a single chart. By, applying different color, it is possible to give a transparent fill or a fill with reduced transparency. This might give a good look for the chart.

unocelli:
Hi, I intend to add the attribute fill in ChartLineZone
```
export interface ChartLineZone {
    min: number;
    max: number;
    color: string;
    fill: string;
}
```

image

Kiruthick-Roxx:
Hi,
   That's really great. Thank you

PR #1539, opened 2024-12-12
Update fr.json

PaoloElMarinelo:
I translate and make corrections to the translation in progress.
I prefer to do several small commits rather than one big one.
I slowly add what's in the English translation so that the `fr` file is the translated reflection of the `en` file.


unocelli:
Good job, Merci beaucoup!

Issue #1572, opened 2025-01-15
[BUG] Cannot read properties of null (reading 'clientX')


CYO9000:
Hey first of all thank you so much for this awesome project! I really enjoy working with fuxa on my projects and never had major issues while doing so. However yesterday i encountered a strange issue when trying to open a card via $opencard in script. Today i tried resolving this issue by installing different versions of fuxa, trying different browsers and looking in code with the help of chatgpt. Unfortunately i wasnt successful and im stuck with this issue. Does anybody have an idea what causes this? Here is an example project which causes this issue on my end:
[test.json](https://github.com/user-attachments/files/18429084/test.json)

**Describe the bug**
When running a script which uses $opencard nothing happens and an error is reported in console.

**To Reproduce**
In the attached project click the button on MainView.

**Screenshots**
![image](https://github.com/user-attachments/assets/b5b6e5d9-d0af-46d6-b8cf-d48998c997fe)

**Environment**
 - FUXA version: 1.2.3-2106
 - Node.js version: 18.20.5
 - npm version: 10.8.2
 - Platform/OS: Windows 10
 - Browser: Chrome/Edge/Firefox

unocelli:
Thanks for the report

CYO9000:
I installed the master-branch and it works as expected thanks a lot!

unocelli:
@mauroalexandre Sorry, I forgot to notify you that I am in charge of it. Thanks.

unocelli:
I’m going to close this as resolved. let me know if you have any issues.

PR #1578, opened 2025-01-17
fix error in clientX and clientY when card is called by script (#1572)

mauroalexandre:


unocelli:
Hi @mauroalexandre, I suppose that we can close the PR?

PR #1589, opened 2025-01-27
Toaster action for Alarms

Kiruthick-Roxx:
One more action to toast the alarms has been added to improve alarm visibility for critical alarms.


unocelli:
Good job, Thanks a lot! let me do the review.

unocelli:
Hi, I did the review and made some changes, the notification loop is not acceptable and I did not understand what spanGaps is for, Please let me know what you think about

Kiruthick-Roxx:
Hi Unocelli,
     Thanks for looking into it. The spangaps toggle button can be used to Turn ON/OFF the fill gap fucntion of Uplot that automatically connects a straight line where no data points available. This feature would help us to customize that so if data is not available, it will be a blank scape and no lines will be plotted there which will be helpful to find out where the data points are missing. Hence, I added that. Please, let me know your suggestions.

PR #1593, opened 2025-01-29
Enable Force Reports for disabled reports

LordNetro:
Remove scheduling type check in reports in order to be able to force a report without having any schedule asociated.

### DIFF
--- a/server/runtime/jobs/index.js
+++ b/server/runtime/jobs/index.js
@@ -142,11 +142,15 @@ function JobsManager(_runtime) {
             runtime.project.getReports().then(function (result) {
                 if (result) {
                     result.forEach(rptProperty => {
-                        if (rptProperty.scheduling !== Report.ReportSchedulingType.none) {
-                            var report = Report.create(rptProperty, runtime);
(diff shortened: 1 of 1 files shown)


unocelli:
Hi, I don't understand the reason, if you disable it is because you don't want them to create reports or send them

LordNetro:
Hi, let me explain. It's because I have the necessity of generating and downloading a report, without having a Schedule defined. I don't need to send it to anyone or generate it after some time. I will execute via API the generate and download calls so from the SCADA you can generate reports on demand, without scheduling mode.

unocelli:
Ok, but then you have to add the endpoint on the server to generate the Report

wonka929:
I agree with the necessity to create reports on demand.
It happens that sometimes a client asks for the report of a particular day (let's say there was a problem on the plant) and right now I have to create it from InfluxDB directly instead of a 1 click report generation from FUXA.
I don't know how it's better implemented, but the option to choose a date and generate the report on the go would be appreciated.

unocelli:
We need a function to be triggered via script

PR #1615, opened 2025-02-21
Feature/vistwo

gh-fx2:
This device allows connection to vistwo scada system.


gh-fx2:
Hi unocelli.
Is there a roadmap when you think to merge my pull request ? Or do you wait for a signal from my side ?
If you need some more informations ....
The new device is using websocket to communicate. There is no polling avail - that's why an own device is needed. For connection only the ip is needed.
Screenshot_device_vistwo
An empty IP-field will interpreted as '127.0.0.1'. The slider 'event driven' will allow to push values into fuxa outside of any polling cycle. ( i.e. if a value is changing 100 times per second - fuxa will paint graph with this speed ). If toggle if 'off' the values will saved inside and only pushed within polling cycle (initiated by fuxa).

Editing the device ...
There is only an address needed to specify the interested data. Items organized in vistwo as a tree (like OPC-UA). Thats why names have slashes...
Screenshot_vistwo_tags

regards
fx2

PS: If you do not plan to merge the pull request - it is okay , but give an short sign , so I can stop merging with your code cyclilly :) .

Issue #1533, opened 2024-12-09
Question: how to use Read Scale Script


jonbur:
Hi, I hope you can help. I am having trouble with Read Scale Script and can't find any documentation.

I have several tags coming from mqtt that hves the value either 'on' or 'off', as below:

<img width="1296" alt="Screenshot 2024-12-09 at 15 23 54" src="https://github.com/user-attachments/assets/8a6fbd8e-fb31-43c3-9644-aae66afb7220">

I would like to use this to drive Actions on a shape. I understand that Actions can only be driven by numerical values and therefore I am looking to convert the text values to equivalent numerical/binary values (1/0). I have created the script below and assigned it as the Read Scale Script.

<img width="395" alt="Screenshot 2024-12-09 at 15 23 25" src="https://github.com/user-attachments/assets/cbca5009-9805-4f58-9959-0142959537fa">

<img width="1082" alt="Screenshot 2024-12-09 at 15 24 32" src="https://github.com/user-attachments/assets/0f153d89-f2d9-488b-8529-62120a7d8b76">

However, this does not appear to work. There is no error in the logs. Even if I change the script to simply just 'return 1' in all cases I don't see a change to the value of the tag. 

Am I missing something?

Thank you very much.

MatthewReed303:
@jonbur at the bottom of the page https://github.com/frangoteam/FUXA/wiki/HowTo-Devices-and-Tags

jonbur:
Hi,

Thanks for this. I have seen this and believe I have followed the design pattern shown, but it does not look like the script is being applied / changing the output. 

Just to make sure I have this correct conceptually, the 'value' that is passed into the read script is the tag value that is read from the integration connector (in this case mqtt) and then the script manipulates it and whatever is returned should then be what we see when that tag is used elsewhere?

Thank you

henjoe:
Your script is actually correct. I dont know why it is not working.

unocelli:
Hi, try to debug with a console.log(value);

jonbur:
Thanks. If I run the script in the script editor with console logging, I get the below:

<img width="1356" alt="Screenshot 2024-12-15 at 11 56 57" src="https://github.com/user-attachments/assets/5d69f38e-418e-4177-985c-cc2ed962244a" />

This seems fine (although I note I don't get return '0' when I call with 'off'...

Is there a logger object I can use to log to fuxa.log from within the script so I can see whether this script is actually getting caleld by the 'read scale script' defined on the tag?

[...]

PR #1643, opened 2025-03-08
fix: the variable of WebAPI encountered an error when executing the Read Scale Script

imdwpeng:
I have found an [issue](https://github.com/frangoteam/FUXA/issues/1533), but the latest code still has problems.
log


unocelli:
Good job, Thanks a lot.

PR #1654, opened 2025-03-18
Feat/pr1637

unocelli:


unocelli:
@rikugun Hi, I have to set the package installation as dynamic, so it has to be installed manually, let me know if the package installed via plugins works, !BE CAREFUL to test it you have to remove it from node_modules so that it only finds it on _pkg.

I still have to do some tests with the hardware, and I saw that the DAQ function is missing

rikugun:
> @rikugun Hi, I have to set the package installation as dynamic, so it has to be installed manually, let me know if the package installed via plugins works, !BE CAREFUL to test it you have to remove it from node_modules so that it only finds it on _pkg.
> 
> I still have to do some tests with the hardware, and I saw that the DAQ function is missing

I've some problem with  installation as dynamic，some module not work with live-plugin-manager, so i change to install manually.  and i'll check DAQ functions.

rikugun:
The package installed via plugins not works.  epoll package can not load with error "Cannot require a module outside a plugin " .
image

image

unocelli:
Ok, but if you install it manually then it works the same way as before? this allows those who want to use GPIO have to install FUXA and then only have to install the onoff package manually (npm -i onoff)

rikugun:
> Ok, but if you install it manually then it works the same way as before? this allows those who want to use GPIO have to install FUXA and then only have to install the onoff package manually (npm -i onoff) 

yes

unocelli:
@rikugun Good job! Thanks a lot!

PR #1637, opened 2025-03-05
Feat/gpio support

rikugun:
add gpio support(linux only). 

image

image

PR #1665, opened 2025-03-26
cleanup: Adaptive client named

rikugun:
#1637 contain bug , naming needs to be consistent


unocelli:
:thumbsup: Thanks a lot!

PR #1666, opened 2025-03-26
feat:  daq query supports group statistics

rikugun:
Provides a query interface for external use and supports group statistics. Reference reporting service


unocelli:
@rikugun it would be great if you explained what it does

rikugun:
I have an external system that calculates the average temperature and humidity at a fixed number of times per day. Like 0, 8, 16 o 'clock sharp

Issue #1293, opened 2024-07-23
[QUESTION] OpenCard multiple instance of View


henjoe:
I created a faceplate view with the following place holders:
1. @TAGNAME
2. @PV
3. @MV
4. @SV 

![image](https://github.com/user-attachments/assets/c9970a1a-9199-4ab4-83b2-e77bde4df240)

On my Main view (shape component), I added some "click" actions with "open card" and bind to the place holder:
![image](https://github.com/user-attachments/assets/fe7c266a-3936-43a9-852f-82d6b1c7d14c)

I did the same on other component on my Main View but this time difference binding of Tags.

What happen is when I open card my Faceplate on the first component, it actually works fine and I have no issue.
![image](https://github.com/user-attachments/assets/7ceda3d4-caf8-4261-ab26-7907f4427f18)

But when I open the 2nd face plate, what happen is it overrides the first Faceplate that is already open. It seems like it is not a new instance of my Faceplate.

![image](https://github.com/user-attachments/assets/1c5d542e-a179-4d7b-b43e-f5e50f79fa04)

As you can see, 2 different binded tags, but they are now identical. 
Is it some sort of bugs or a normal behaviour?

Regards,

robsori:
Hi. It's an older problem. See here: https://github.com/frangoteam/FUXA/issues/744

henjoe:
I did it!!! Almost 2 days of debugging and coding. 
I will test it more first then will do the PR.

![image](https://github.com/user-attachments/assets/2a50d53f-de2d-4d15-87c6-8b61753cf4e0)

unocelli:
#308
#805

henjoe:
@unocelli I already have working program for this. 
The concept is to add contructor on the view, and replace the view id of the same view so that it doesn't conflict and it can open multiple instance of Card (Popup) view. 

The main refactor is in the View constructor, and in fuxa component openCard.  I can share you the code and simply resolve conflict, sorry not really familiar with the PR thing,  (the last time I check, the latest fuxa is totally different from what I have now), but if you have any questions I can answer it.

unocelli:
@henjoe It would be great if you could share with me how you solved it, when I analysed it it seemed quite challenging to solve it.

henjoe:
Hi @unocelli  please check PR #1687

PR #1687, opened 2025-04-12
Feature/open card multi instance

henjoe:
Allows to open multi-instance of open Card. (just like a typical faceplate)  #1293 

image


unocelli:
Pity, the feature will have to be addressed sooner or later

PR #1700, opened 2025-04-19
Update hmi.service.ts

carlbomsdata:
When $setView() was called it opened up a view but also a card. Now it only open up a view.


unocelli:
Thanks a lot

PR #1826, opened 2025-06-24
Fix: allow wildcard IP patterns in CORS origin check  issue #1816

### LINKED ISSUE #1816: [BUG] settings.js
The CORS standard does not allow wildcard matching like http://192.168.*. The browser requires an exact match for the origin in the Access-Control-Allow-Origin header.

line 64 
```
"allowedOrigins": ["http://localhost", "http://127.0.0.1", "http://192.168.*", "http://10.*"]
```

manish-kumar-developer:
Browsers require exact origin match in Access-Control-Allow-Origin. Replaced Array.includes with regex matching to support patterns like http://192.168.* and http://10.* in development environments. "

issue fixed #1816

### DIFF
--- a/server/main.js
+++ b/server/main.js
@@ -296,14 +296,6 @@ const allowCrossDomain = function (req, res, next) {
     return res.sendStatus(204); 
   }
   next();
-  
-  try {
-    const ip = req.headers['x-forwarded-for'] || req.connection.remoteAddress;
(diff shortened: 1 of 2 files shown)


manish-kumar-developer:
yes it was just for confirmation for developer there is no such specific reason/dependency we can remove it

Issue #1776, opened 2025-05-26
[Question] Force using function code 16 (Write Multiple Register)


steve0023:
Hello, is there any way to force to use the function code 16 (Write Multiple Register) instead of the function 6 (Write Single Register) to write a single register in ModbusTCP ?
Unfortunately my PLC doesn't manage function 6 ....

Thanks.

unocelli:
I assume the issue has been resolved in the meanwhile.

PR #1862, opened 2025-07-16
Feature : Selector To use Fonction Write Multiple Registers when writing 1 register

steve0023:
This feature introduces a selector that allows the use of Function Code 16 (Write Multiple Registers) instead of Function Code 6 (Write Single Register), even when writing a single register.

I'm not very experienced with JavaScript, but the code works as intended. I hope it meets the required standards—please feel free to modify it as needed.

This addresses issue #1776.


steve0023:
Did somes mistake on by branch, git newbee...

PR #1982, opened 2025-10-21
OPC UA - UInt32 Syntax Error

LordNetro:
There is a syntax error on the opcua device regarding UInt32.


unocelli:
@LordNetro Thanks a lot!

Issue #211, opened 2021-10-28
Modbus RTU multiple devices per serial port.


T1bbY:
Hello,

when i create multiple device's as ModbusRTU and want to use the same serial port, it does not work.
That make sense as it try's to open a new connection to the serial port.

But i habe a ModbusRTU with multiple devise's and all different ID's and i can only access one at a time.

Would be nice to add the possibility to communicate with multiple ID's on the same RTU connection.

Kindly regards :)

zhang9591:
Hope to see this update！

rvbatista:
I studied the code a little bit and seems that need another layer for the physical  connection (serial port) above the device layer, or assumes that the device is the serial port and each tag have the modbus unit id property. The same will apply to issue #755.

unocelli:
@rvbatista Hi, I have not yet examined it well, but I think it might be sufficient to map all tags of all devices configured as RTUs to one device by grouping them by Id.

rvbatista:
@unocelli this will be faster to do, but will break the timeout and connect logic. I'm not experienced in node and zero experience in angular, but it's I'm interested in trying to implement this modification.

unocelli:
@rvbatista polling waits for the response with await and before reading you can sets the respective deviceId, after correct mapping you extend _readMemory with the deviceId parameter and before reading it you sets the id `client.setID(deviceId);`

unocelli:
@rvbatista thinking back, maybe it is simpler to share the client component (ModbusRTU) between the different devices and define a main device (the first one) that handles the connection, in this case you don't have to make a new mapping but to synchronize the communication

PaDaBooo:
I implemented the function of multiple SlaveID under ModbusRTU by adding a SlaveID property to each Tag.
Do you have a better solution?
If not, I will submit the server code tomorrow.

PaDaBooo:
By the way, sharedDevices in src don't seem to be useful
![image](https://github.com/frangoteam/FUXA/assets/133078679/afaaac4c-1897-4cde-b3d6-319085bc3730)

rvbatista:
> By the way, sharedDevices in src don't seem to be useful ![image](https://user-images.githubusercontent.com/133078679/268044605-afaaac4c-1897-4cde-b3d6-319085bc3730.png)

@PaDaBooo this sharedDevices was the beginning of development and really does nothing at this moment. I didn't had much time to work with this issue. The solution that you implemented is not the best, but better to have something as work around than way for the best solution.
[...]

PR #2000, opened 2025-10-28
Modbus RTU multiple devices in the same serial port.

rvbatista:
Relate to issue #211
It was applied the same solution as the Multiple devices on modbus TCP.


unocelli:
@rvbatista Thanks a lot for the update.

rvbatista:
Sorry for the huge amount of changed lines with only a space.

PR #2016, opened 2025-11-07
Add heartbeatIntervalSec setting for server heartbeat configuration

unocelli:
### ✨ Description
This PR introduces a new setting heartbeatIntervalSec in server/_appdata/settings.js.
It defines how often the server sends heartbeat signals to connected clients (in seconds).
The value can be configured between 1 and 20 seconds, with a default of 10 seconds.

If the client does not receive a heartbeat within 30 seconds, it automatically displays the banner:

SERVER CONNECTION FAILED!

### Changes:
- Added heartbeatIntervalSec to settings (undefined default: 10 s)
- Used to control the periodic server heartbeat toward clients
- 30 s delay before showing the disconnection banner

### DIFF
--- a/client/src/app/home/home.component.ts
+++ b/client/src/app/home/home.component.ts
@@ -128,7 +128,7 @@ export class HomeComponent implements OnInit, AfterViewInit, OnDestroy {
                 switchMap(([connectionStatus, userProfile]) =>
                     merge(
                         of(false),
-                        timer(20000).pipe(map(() => (this.securityEnabled && !userProfile) ? false : true)),
+                        timer(35000).pipe(map(() => (this.securityEnabled && !userProfile) ? false : true)),
                     ).pipe (
                         startWith(false),
                     )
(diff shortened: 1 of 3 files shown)


aiguesmanresa:
Hello @unocelli, it seems the 10 seconds interval is not being actually applied, unless I'm misunderstanding this. I have this in settings.js
```
heartbeatIntervalSec: 10,
```
But the heartbeat is sent every 5 minutes, as defined in client/src/app/_services/heartbeat.service.ts. I've attached a har file:
[10.10.11.20_Archive [26-01-02 12-25-00].zip](https://github.com/user-attachments/files/24406930/10.10.11.20_Archive.26-01-02.12-25-00.zip)

unocelli:
@aiguesmanresa Hi, the server send the heartbeat to client, you will find the code in `FUXA\server\runtime\index.js` line 350

`
    setInterval(() => {
        io.emit(Events.IoEventTypes.ALIVE, { message: 'FUXA server is alive!' });
    }, (settings.heartbeatIntervalSec || 10) * 1000);
`
the client if not receive the heartbeat from server show the banner SERVER CONNECTION FAILED!

the heartbeat from client to server is used to check the token and refresh it

PR #2040, opened 2025-11-20
feat: Using setValue to update UI immediately for WebAPI devices

### LINKED ISSUE #1590: [QUESTION]: websocket usage for data updates
Hi there,

I’m experiencing an issue when trying to use the new WebSocket option in FUXA to update data more frequently than the 1-second interval provided by the WebAPI.

Here’s what I’ve done:

1. 	Switched my WebAPI connection to an “internal” connection to retain the tags.
2. 	Created a server-side script to fetch the initial data using the WebAPI.
3. 	Set up the WebSocket connection to handle subsequent updates.

However, the behavior isn’t as expected. When fetching tag data by ID using `$getTag()`, it always returns `null`.

Example API Response:

From my WebAPI:

```
{
    "t_8f60fd-f8073b": "1",
    "t_cb52db-724556": "3",
    "t_a2d094-5596cc": "1"
}
```

The WebSocket returns the following:

```
42/fuxa,
[
    "fuxa-data",
    {
        "tagId": "t_9bece6-11c1c5",
        "data": "-21.2713623046875"
    }
]
```

When I try to then `setTag(tagId, data)` I always get `setValue not supported!`

Despite successfully retaining tag IDs and verifying their values using the template function `$getTag()`, attempting to access tags via `$getTag('t_8f60fd-f8073b')` consistently returns `null`.

`console.log($getTag('t_8f60fd-f8073b' /* stHMIInterface - iCurrentUnit */));`

Environment:
FUXA Version: `1.2.3`

Expected Behavior:

- Only have an internal device with tags. 
- `$setTag` to set the data to a internal device tag to get rid of the webapi connection

Please let me know if I’m missing something in the setup or if this is a bug. Any guidance would be much appreciated!

Thank you!

Script that I configured as 1s startup delay on server: 

```
const readDataFromWebsocket = () => {
  // Handle successful connection
  ws.on('open', () => {
    console.log('WebSocket connection established');
    ws.send('40/fuxa');
  });

  ws.on('message', (message) => {
     try {
        const messageStr = message.toString();
        if (messageStr.startsWith('42/fuxa,')) {
          const parsedData = JSON.parse(messageStr.slice(8)); // Strip off the socket prefix "42/fuxa"
          const tagId = parsedData[1].tagId;
          const value = parsedData[1].data;
          console.log('tagId', tagId);
          console.log('value', value);
          $setTag(tagId, value);
        } else {
          console.log('Raw message:', messageStr);
        }
      } catch (error) {
        console.log('Error processing message from WebSocket server:', error);
      }
  });

  ws.on('error', (error) => {
    console.log('WebSocket error:', error);
  });

  ws.on('close', () => {
    console.log('WebSocket connection closed');
    //TODO: Attempt to reconnect after a delay if the connection is closed
  });
}

const getInitialFetchData = () => {
	const options = {
    hostname: hostname,
    port: port,
    path: getPath,
    method: 'GET',
  };

  const req = http.request(options, (res) => {
    let responseData = '';

    res.on('data', (chunk) => {
      responseData += chunk;
    });

    res.on('end', () => {
      const data = JSON.parse(responseData);
      if (data) {
        Object.entries(data).forEach(([tagId, value]) => {
           $setTag(tagId, value);
        });
        
        readDataFromWebsocket()
      } else {
        console.log('empty data');
      }
    });
  });

  req.on('error', (error) => {
    console.log('Error');
    console.log(JSON.stringify(error));
  });

  req.end();

}

getInitialFetchData()

```

ShadowBr0ther:
Implements `setValue` functionality for WebAPI devices, enabling immediate UI updates when values are changed programmatically via WebSocket or scripts, bypassing the polling interval.

### Problem

WebAPI devices did not support `setValue`. When updating tag values programmatically through WebSocket commands or scripts, the UI would not reflect changes until the next polling cycle, causing delays of several seconds depending on the polling interval.

This made real-time interactions (e.g., script runs for calling external actions) feel unresponsive like in the case of issue #1590

### Use Case Example

A **WebAPI device** is configured with a polling interval (e.g., 5 seconds) to regularly fetch data. Additionally, a **WebSocket connection** is established with a script (similar to the WebSocket integration described in the FUXA wiki [HowTo Websockets](https://github.com/frangoteam/FUXA/wiki/HowTo-WebSockets)) that receives real-time push notifications from the WebAPI server when values change.

  **How it works:**

  1. **WebSocket pushes update** → Server sends real-time value change notification
  2. **`setValue()` is called** → FUXA processes the update and immediately refreshes the UI
  3. **Polling continues** → Regular polling runs in the background as a fallback to guarantee data consistency
 
 **Benefits of the possibility  of this hybrid approach:**

  -  **Real-time responsiveness**: UI updates instantly via WebSocket push notifications
  -  **Guaranteed reliability**: Polling acts as a fallback mechanism to catch any missed WebSocket messages
  -  **Best of both worlds**: Low latency when possible, with guaranteed eventual consistency

  ### Related Issues

  Fixes #1590

### DIFF
--- a/server/runtime/devices/httprequest/index.js
+++ b/server/runtime/devices/httprequest/index.js
@@ -129,7 +129,7 @@ function HTTPclient(_data, _logger, _events, _runtime) {
      * Return the timestamp of last read tag operation on polling
      * @returns
      */
-     this.lastReadTimestamp = () => {
+    this.lastReadTimestamp = () => {
         return lastTimestampValue;
     }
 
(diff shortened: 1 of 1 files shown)


unocelli:
@ShadowBr0ther Hi, I don't understand the scope to update the UI value without post to device, you will see a wrong information that don't reflect the device value.

ShadowBr0ther:
Hi @unocelli,

you’re absolutely right that blindly updating the UI without writing to the real device can lead to inconsistent data. I don’t want to argue against that in general – the point of this change is a bit more specific.

In our concrete WebAPI + WebSocket setup it works like this:

- The WebAPI device is a heavy pull endpoint (large JSON, costly backend processing, etc.).
- In parallel, we have a WebSocket connection from the same backend that pushes authoritative value changes in real time.
- When such a WebSocket message arrives, we call setValue() so that FUXA updates the UI immediately, instead of waiting for the next poll.

So we are not inventing values in the frontend and we’re not bypassing the real device logic. The value has already changed on the backend / device side – the WebSocket is just the faster notification channel. Polling stays enabled as a fallback, so eventual consistency is still guaranteed even if a WebSocket message is lost.

Without this behavior, to achieve the same responsiveness we’d have to:
- drastically lower the polling interval (which would hammer the WebAPI endpoint), or
- accept several seconds of lag between a real device change and the UI.
Both of this options are real pain points.

With this hybrid model:
- Real-time responsiveness comes from WebSocket pushes + setValue() updating the UI,
- Reliability comes from the existing polling that keeps the state in sync over time
- The values themselves are identical – just delivered more efficiently and with lower latency.

Of course this can be misused (e.g. scripts setting arbitrary fake values), but that requires someone to explicitly implement that misuse in their script. For legitimate WebAPI/WebSocket setups the feature simply exposes what the backend already knows, faster and with much less load on the infrastructure.

unocelli:
Hi @ShadowBr0ther Thanks for the PR. The goal makes sense, but I think this needs some caution.

For `ownFlag=false`, `setValue` now returns success, updates local `varsValue`, emits to the frontend and may write to DAQ (Chart), but it does not actually write anything to the external API. That can be misleading, because the value can be overwritten by the next polling cycle.

This also affects restore: `restoreValues()` calls `setValue()` on startup, so read-only WebAPI devices may temporarily restore stale local values and emit/save them before the real polling value arrives.

A safer approach could be adding an explicit WebAPI device flag, e.g. `allowLocalSetValue`, disabled by default. When enabled, `setValue` can perform this local/frontend update for `ownFlag=false`; otherwise the current read-only behavior is preserved.

PR #2063, opened 2025-12-03
Feat/ODBC Browser, History Table ODBC, New Parameter Table

MatthewReed303:
## ODBC Browser and Enhanced Table Features

This PR introduces comprehensive ODBC database integration and enhanced table functionality to FUXA SCADA.

### Key Features Added:

**ODBC Browser**
- New ODBC browser component for database exploration and query building
- Support for browsing database schemas, tables, and columns
- Query builder service with SQL syntax assistance
- Column definition analysis (types, constraints, relationships)

**Database Integration**
- ODBC driver handling and database connection support
- Date formatting utilities for ODBC queries
- Enhanced DB support with proper data type handling

**New Parameter Table Type**
- Complete parameter table implementation with full backend support
- CRUD operations for parameter management
- SQLite and ODBC datasource integration
- Auto-write functionality on application startup

**Table Enhancements**
- Improved history table with chunked data loading for large datasets
- Hash-based data integrity checks
- Enhanced UI with better styling and user experience
- Optimized large data handling and auto-refresh capabilities

### Technical Changes:
- Added ODBC device runtime support for different database types
- New API endpoints for parameter table management
- Enhanced table components with advanced features
- Improved data persistence and caching mechanisms


concept10:
500,000 LOC on this PR. Wow (ridiculous IMO no matter the added functionality) but that's just me.

unocelli:
Nice work on this feature — it’s a pretty big addition 👍  

As a first pass, I’d fix a couple of high-level things before we go deeper:

## 1. Build artifacts  
Please remove the generated build files from the PR.

## 2. Security (ODBC / DB access)  
Since this introduces quite a bit of database and ODBC logic, we should double-check the security side — especially how queries are handled on the server.

Where possible, we should prefer parameterized queries instead of building SQL strings dynamically. For example:

```js
const query = `SELECT * FROM users WHERE id = ${userId}`;
```

would be safer as:

```js
const query = `SELECT * FROM users WHERE id = ?`;
const result = await connection.query(query, [userId]);
```

This way the driver handles escaping and we reduce injection risks.
For dynamic identifiers (like table names), we should make sure they are resolved server-side from a controlled source rather than trusting raw client input.

## 3. ODBC device access

The API lets the client choose the storage type and ODBC device. We should make sure this follows the same authorization pattern used in similar endpoints — i.e. enforce the check when authorization is configured, just like we do elsewhere.

Once this is clarified/adjusted, we can go through the implementation in more detail.

MatthewReed303:
@unocelli have done all the above and fixed Angular v18 issues and merge conflict issues

unocelli:
@MatthewReed303 Great, Please remove dist folder from PR.

MatthewReed303:
@unocelli sorry removed them with commit but not from the commit history, all done now

MatthewReed303:
> 500,000 LOC on this PR. Wow (ridiculous IMO no matter the added functionality) but that's just me.

@concept10 this was due to build files included, all cleaned up now

unocelli:
Hi @MatthewReed303 

I prepared a separate branch with the changes I tested locally instead of pushing directly to your PR branch.

Branch:
`fix/odbc-browser-review`

Main commit:
`338f5caf` (`fix: ui5 component removed`)

What I changed / why:

I removed the UI5 integration from the ODBC browser.

The UI5 webcomponents addition looked too expensive for the scope of this feature:
- it introduced a second UI stack into an Angular Material app
- it required `CUSTOM_ELEMENTS_SCHEMA` at app level, which weakens Angular template validation globally
- it added extra package/runtime/theming/localization complexity for a very narrow use case
- it was also breaking the client build in my local test until removed

I replaced the UI5 date/time inputs with standard Angular-compatible inputs (`date`, `time`, `datetime-local`) inside the existing Material form fields. That keeps the feature simpler and avoids the extra dependency cost.

Testing notes:

I tested the PR with both PostgreSQL and SQL Server ODBC connections.

What I observed:
- the ODBC connection itself can work
- the device connection dialog can retrieve table names
- the browser/query/data viewer flow is still not reliable
- PostgreSQL and SQL Server both still show functional issues in the ODBC Browser path
- because of that, I reverted my local connection-path fix and I think this part should be solved properly in the PR first

So my current view is:
- the feature direction makes sense
- the UI5 addition does not seem worth the cost here
- the connection / browser / data viewer flow still needs to be stabilized before merge

If useful, you can compare or cherry-pick from `fix/odbc-browser-review`.

MatthewReed303:
@unocelli I will work on your new branch. As for connection path issues, what exact issues did you have? I tested on multiple database types with no issues ( my dev setup is on Linux )

unocelli:
@MatthewReed303 I'm having the problem described on Windows

PR #2070, opened 2025-12-05
Add retension for logs

MatthewReed303:
# Add Log Retention Settings

## Overview
This PR adds a new "Logs" tab to the FUXA settings dialog, allowing users to configure automatic cleanup of old log files based on retention periods, similar to the existing DAQ and Alarms retention features.

## Changes
- **Client**: Added LogsSettings model, updated settings component and template to include a new "Logs" tab with retention dropdown (Disabled, 1 day, 2 days, ..., 5 years).
- **Server**: Implemented log cleanup logic in the daily cleaner job, deleting log files older than the selected retention period.
- **Settings**: Added logs retention to default settings and user settings persistence.

## Features
- Retention options match DAQ settings for consistency.
- Default is "Disabled" (keeps all logs).
- Automatic cleanup runs daily as part of the existing cleaner process.
- Cross-platform compatible (Node.js fs module).

### DIFF
--- a/client/src/app/_models/settings.ts
+++ b/client/src/app/_models/settings.ts
@@ -17,6 +17,8 @@ export class AppSettings {
     daqstore = new DaqStore();
     /** Alarms store settings */
     alarms = new AlarmsSettings();
+    /** Logs settings */
+    logs = new LogsSettings();
     /** Log Full enabled to log all setValue */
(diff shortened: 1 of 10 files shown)


unocelli:
@MatthewReed303 Good job!

PR #2089, opened 2025-12-13
feature: fix for mobile scaling

tobias-carlbom:
Before, init scale was always set to 1. now the init scale is calculated. 

On orientation change we reload the ui.

Also, pitch zoom is disabled on mobile/tablets. Zooming inside Fuxa view still works.


unocelli:
@tobias-carlbom Good job, Thanks a lot

PR #2116, opened 2026-01-06
Fix modbus rtu save socketreuse

rvbatista:
fix device modbus rtu save socketreuse


unocelli:
@rvbatista Thanks a lot!

Issue #2128, opened 2026-01-11
Add 50ms and 100ms polling options


trlaw:
Add 50ms and 100ms polling options for connected devices

Desired so can use Fuxa for machine mock-up animations for design meetings

PR #2129, opened 2026-01-11
Add 50ms and 100ms polling rate options

trlaw:
Tested locally and was functional for creating 'light' animations using OpenPLCruntimeV3

Closes Issue #2128


unocelli:
Thanks for the contribution! 
I’ll add a warning note to clearly mark these values as advanced options, since they may increase CPU and network load.

PR #2132, opened 2026-01-15
Fixed unclear 'Not Supported!' errors

Delorean12DMC:
Added advanced output to the 'Not Supported!' errors in the index.js files to better understand where the error comes from.


unocelli:
Thanks for the improvement!

PR #2141, opened 2026-01-18
feat(node-red): enhance tag handling with optional tagId for backward compatibility

rikugun:
sometimes, we have the same tag name and device name, but getTagId return first tag.

- Added a new optional field `tagId` to various nodes for unique identifier support.
- Updated UI elements to allow selection of tags in the format `tag.name(tag.id)`.
- Improved backward compatibility by allowing nodes to function with either `tag` or `tagId`.
- Adjusted error messages to reflect the new tag handling logic.


unocelli:
@rikugun Hi, thank you very much for the PR

I really appreciate the work you’ve done.
@MatthewReed303 can you take a look please?

Thanks again for the contribution! 🚀

unocelli:
@rikugun Thanks a lot!

PR #2142, opened 2026-01-18
feat: add scroll buttons to logs view

rikugun:
Add scroll buttons at logs view  right bottom

image


unocelli:
@rikugun Good job, thanks!

PR #2151, opened 2026-01-22
Morgan disable

Delorean12DMC:
Added option to completely disable HTTP request logging to clean the console for custom debugging scripts and to reduce I/O overhead.


unocelli:
Good jobs, Thanks!

PR #2214, opened 2026-02-24
Fix/daqstorage unhandled promise rejection

zarch:
## 📌 Description

Problem: The DAQ storage layer emitted recurring UnhandledPromiseRejection errors paired with daqstorage.data-overload warnings on every poll cycle when a device tag was being stored.

Two bugs in `server/runtime/storage/sqlite/index.js` caused this:

1. _insertTagValue called reject() with no argument, making the rejection reason undefined instead of the actual SQLite error.
2. In addDaqValues, the _insertTagValue promises were created before the concurrency guard (_checkDataWorking) was checked. When the guard returned false (overload), those already-running promises had no rejection handler, causing Node.js to raise UnhandledPromiseRejection.

Changes:
- _insertTagValue now calls reject(err) so the rejection reason is the actual SQLite error.
- addDaqValues attaches no-op .catch() handlers to orphaned promises when the overload guard fires.
- Regression tests added in server/test/storage/daqnode.test.js to cover both cases.

---

## 🧪 Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Other

---

## 🚫 Build Artifacts Check

Please confirm:

---

## 🔍 Checklist

---

## 📝 Additional Notes

The issue manifests as a pair of log lines repeating on every poll cycle (typically every 1–3 s):

```log
[warn] daqstorage.data-overload! '0'
[error] UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "undefined".
```

The regression tests verify both failure modes: running npm test with the fix reverted reproduces the failures, and with the fix applied all tests pass.

### DIFF
--- a/server/runtime/storage/sqlite/index.js
+++ b/server/runtime/storage/sqlite/index.js
@@ -227,14 +227,14 @@ function DaqNode(_settings, _log, _id, _currentStorate) {
                         _checkDataWorking(false);
                     });
                 } 
-                // check function to add daq data            
+                // check function to add daq data
                 if (addDaqfnc.length > 0) {
                     if (_checkDataWorking(true)) {
                         Promise.all(addDaqfnc).then(result => {
                             // check db tokenizer after inserted
                             var lastts = 0 || result[0];
                             if (daqNextToken && daqNextToken < lastts) {
-                                // close data DB, open and bind the new db, rename and move the closed db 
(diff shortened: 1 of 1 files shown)


unocelli:
@zarch Good job, Thanks a lot!

PR #2236, opened 2026-03-05
feat: improved layered dockerfile and dockerignore

alext-extracellular:
## 📌 Description

This PR refactors the dockerfile in order to build a much smaller image (2.4gb -> 540 MB), and to also make use of build layer caching for faster repeated builds. 

The git repository is no longer cloned into the image and it is assumed it is being built from within the repo. 
This allows building from different tags and branches as well, and should still work in CI.

This is needed because the image was way too large before and took ages to build.

---

## 🧪 Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactoring
- [ ] Documentation
- [ ] Other

---

## 🚫 Build Artifacts Check

Please confirm:

---

## 🔍 Checklist

---

- [ ] All links were tested

---

## 📝 Additional Notes

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,30 +1,7 @@
-# Dependency directories
-# We want the container to run its own 'npm install'
-node_modules/
-npm-debug.log
-yarn-error.log
-
-# Git metadata
 .git
-.gitignore
-
-# Local build artifacts and caches
-dist/
-build/
-.cache/
-.npm/
-
-# OS-specific files
-.DS_Store
-Thumbs.db
-
-# Docker-related files
-Dockerfile
-.dockerignore
-docker-compose.yml
-
-# Large data or temp files
-*.tar
-*.zip
-*.log
-tmp/
+.angular
+**/node_modules
+**/dist
+app/electron
+docs
+screenshot
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,30 @@
+# Dependency directories
+# We want the container to run its own 'npm install'
+node_modules/
+npm-debug.log
+yarn-error.log
+
+# Git metadata
+.git
+.gitignore
+
+# Local build artifacts and caches
+dist/
+build/
+.cache/
+.npm/
+
+# OS-specific files
+.DS_Store
+Thumbs.db
+
+# Docker-related files
+Dockerfile
+.dockerignore
+docker-compose.yml
+
+# Large data or temp files
+*.tar
+*.zip
+*.log
+tmp/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,57 +1,78 @@
-# --- STAGE 1: The Builder ---
-FROM node:18-bookworm AS builder
+# --- STAGE 1: Angular Client Builder ---
+FROM node:18-bookworm AS client-builder
+WORKDIR /usr/src/app/client
+COPY client/package*.json ./
+RUN npm install --no-audit --no-fund
+COPY client/ ./
+RUN npm run build -- --configuration production
 
-ARG NODE_SNAP=false
+# --- STAGE 2: Server & Native Dependencies Builder ---
+FROM node:18-bookworm AS server-builder
+# Define build arguments with defaults
+ARG INSTALL_SNAP=false
+ARG INSTALL_ODBC=true
+
 WORKDIR /usr/src/app/FUXA
 
-# Install ONLY what is needed to compile
+# Base build tools (always needed for SQLite)
 RUN apt-get update && apt-get install -y \
-    dos2unix build-essential unixodbc-dev sqlite3 libsqlite3-dev \
+    python3 build-essential libsqlite3-dev dos2unix \
+    $( [ "$INSTALL_ODBC" = "true" ] && echo "unixodbc-dev" ) \
     && rm -rf /var/lib/apt/lists/*
 
-# Copy only dependency files first to leverage caching
+# Install Server dependencies
 COPY server/package*.json ./server/
 WORKDIR /usr/src/app/FUXA/server
+RUN npm install --no-audit --no-fund
 
-# Optimized NPM for speed
-RUN npm config set audit false && npm config set fund false
-RUN npm install --no-audit --no-fund --network-timeout=600000
+# Optional Snap7 installation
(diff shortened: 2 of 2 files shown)


unocelli:
Hi @alext-extracellular Thanks for the contribution.

I tested the Docker build locally and the resulting image size is around 7.3GB, which is significantly larger than the current image (~4GB).

It seems that some large directories might still be copied into the image. Could you double-check the .dockerignore and the Dockerfile layers?

Once the image size is actually reduced we can review it again.

alext-extracellular:
Wow that is quite large, yes probably some differences in what is in our personal repos, I will try to fix that.
Here is my built image:
```
frangoteam/fuxa:testsmall4                             6494f4d4f340        578MB             0B
```
Here is a depth 2 tree of my repo:
```
.
├── app
│   └── electron
├── client
│   ├── dist
│   ├── e2e
│   ├── node_modules
│   ├── src
│   ├── angular.json
│   ├── bs-config.json
│   ├── package-lock.json
│   ├── package.json
│   ├── proxy.conf.json
│   └── tsconfig.json
├── docs
│   ├── assets
│   ├── images
│   ├── Getting-Started.md
│   ├── HowTo-animate-Pipe.md
│   ├── HowTo-bind-Controls.md
│   ├── HowTo-bind-Shapes.md
│   ├── HowTo-Chart-Control.md
│   ├── HowTo-configure-events.md
│   ├── HowTo-configure-Script.md
│   ├── HowTo-define-Shapes.md
│   ├── HowTo-Devices-and-Tags.md
│   ├── HowTo-Node-Red.md
│   ├── HowTo-ODBC.md
│   ├── HowTo-save-load-Project.md
│   ├── HowTo-Scheduler.md
│   ├── HowTo-setup-Alarms.md
│   ├── HowTo-UI-Layout.md
│   ├── HowTo-use-same-view.md
│   ├── HowTo-View.md
│   ├── HowTo-WebSockets.md
│   ├── HowTo-Widgets.md
│   ├── index.md
│   ├── Installing-and-Running.md
│   ├── Settings.md
│   └── Tips-and-Tricks.md
├── node-red
│   └── node-red-contrib-fuxa
├── node_modules
│   ├── @serialport
│   ├── async-mutex
│   └── tslib
├── odbc
│   ├── install_odbc_drivers.sh
│   └── odbcinst.ini
├── screenshot
│   ├── feature-action-move.gif
│   ├── fuxa-device.gif
│   ├── fuxa-editor.png
│   ├── fuxa-hmi.gif
│   ├── fuxa-home.png
│   ├── fuxa-test.gif
│   └── fuxa-thinglinks.gif
├── server
│   ├── _images
│   ├── _widgets
│   ├── api
│   ├── dist
│   ├── docs
│   ├── integrations
│   ├── node_modules
│   ├── runtime
│   ├── test
│   ├── envParams.js
│   ├── fuxa.js
│   ├── main.js
│   ├── package-lock.json
│   ├── package.json
│   ├── paths.js
│   ├── project.default.json
│   ├── project.demo.fuxap
│   ├── settings.default.js
│   └── tsconfig.json
├── bakDockerfile
├── betterfile
├── CHANGES.md
├── compose.yml
├── CONNECTION_MANAGER.md
├── CONTRIBUTING.md
├── Dockerfile
├── LICENSE
├── mkdocs.yml
├── modbus-rewrite.md
├── package-lock.json
├── package.json
├── README.md
├── SECURITY.md
└── TAG_LEVEL_SLAVE_ID_IMPLEMENTATION.md
```

@unocelli If you can, it would be handy to see a tree of ur repo

alext-extracellular:
OK, I have updated the dockerfile and dockerignore. I have tested by cloning a fresh repo and running the build, the output is below:

```
frangoteam/fuxa:latest        3701395eae46       1.43GB             0B   U
fuxa_test:latest              91ef9e2de901        587MB             0B
```
Make sure you are using the .dockerignore, otherwise some large stuff might get copied into the output.

unocelli:
Hi @alext-extracellular I re-checked the image size on a clean build, and the overall direction of this PR looks good now. In my test, the current origin/master image was about 1.23 GiB, while this PR built an image of about 183 MiB, so the size reduction is real and significant.

There are just two small functional points I think would be worth keeping aligned with the previous behavior before merge:

1. When NODE_SNAP=true, it would be good to make sure node-snap7 is still guaranteed to be present in the final runtime image. Right now it is installed outside server/package.json, and then the final stage runs npm prune --production, so I think there is a chance it may get removed there.

2. The Dockerfile now removes server/docs, but the server still reads server/docs/openapi.yaml when Swagger is enabled. It may be safer to keep that file in the runtime image so this case continues to work as before.

Other than that, the image optimization itself seems to be working well now.

alext-extracellular:
@unocelli OK, in this latest commit I have made sure that node-snap7 ends up in the final build if the arg is set. Instead of pruning after, we prune before and then optionally install snap7.

Also stopped it deleting the openapi docs.

unocelli:
Hi @alext-extracellular looks good. Thanks a lot!

wsfyh666:
"That’s really impressive — I truly appreciate the great work from both of you! @unocelli @alext-extracellular 
I have one suggestion regarding security hardening:
To follow the principle of least privilege, would it be possible to configure the container to run as a non-root user? Running as root is generally discouraged in production environments to minimize the risk of container escape vulnerabilities.
Thanks again for the great work!"

alext-extracellular:
Hi @wsfyh666 yes it is a good idea, however I wonder if it may break connections that require a physical device like modbus rtu. As normally in linux those can only be opened by root. However it may be possible to use udev or perhaps add to dialout group. but I think we have to be very careful not to cause a breaking change.

One idea is to run only the startup script (entrypoint.sh for example) as root, scan for devices, then add them to udev, then launch fuxa as non-root. but this would then not allow dynamic addition of devices (if the user mounts /dev:/dev in the container, which I need to do often)

Additionally there is now node-red being used inside fuxa, and whilst that adds complexity and attack surface, it also means that there is greater room for breaking changes if we change the permissions. Since anyone could be doing practically anything with node red, we can't really know what will break by changing perms.

@unocelli do you have any thoughts?

wsfyh666:
Hi @alext-extracellular Thanks for the detailed explanation — I really appreciate your perspective and it makes a lot of sense given the existing deployments and hardware constraints. Also thanks for your work on the project!

PR #2245, opened 2026-03-11
Fix/project import 413 error feedback

zarch:
## 📌 Description

Problem: Importing a large project file (≥ 184 MB in the reported case) through the FUXA UI failed silently. The interface showed only a generic notification with `Project save failed!` with no indication of the actual cause, and the server logs contained no application-level error entry — only the raw HTTP access log line `POST /api/project 413`.

Three separate issues combined to produce this behaviour:

1. **Server silent on 413**: In `server/api/index.js` the Express error handler for `entity.too.large` returned the 413 response but never called `runtime.logger.error()`. The rejection was invisible in application logs.

2. **Frontend discarded HTTP status**: In `client/src/app/_services/project.service.ts`, the `save()` method had its own inline error handler that always showed the translated `msg.project-save-error` string, ignoring the HTTP status code entirely. The existing `notifySaveError()` helper already handled 401 and 413 correctly (surfacing the server message for 413, and a hint for status 0 / connection abort), but `save()` bypassed it.

3. **Angular client could not be rebuilt**: `client/src/theme.scss` used `mat.m2-define-*` functions and `mat.$m2-*` palette variables introduced in Angular Material 17, while the installed package is 16.2.13. This caused an "Undefined function" Sass compilation error that prevented any rebuild of the client bundle.

Changes:
- `server/api/index.js`: added `runtime.logger.error()` in the 413 error handler so the rejection appears in application logs alongside the HTTP access log entry.
- `client/src/app/_services/project.service.ts`: replaced the inline error handler in `save()` with a call to `notifySaveError(err)`. Added a `status === 0` branch in `notifySaveError()` for browsers that abort the connection mid-upload rather than receiving the 413 cleanly. Extended `ServerSettings` with `apiMaxLength` so the configured limit is available for that message.
- `client/src/theme.scss`: replaced all `m2-` prefixed Sass identifiers with their Angular Material 16 equivalents (`mat.define-*` / `mat.$*-palette`) to restore the ability to build the client. I keep these change in a separate commit because I do not know if you are in a transition phase from Angular Material 16 to Angular Material 17, but without this fix I was not able to test it locally. So in case let me know and I rebase it without this last commit.

---

## 🧪 Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Other

---

## 🚫 Build Artifacts Check

Please confirm:

---

## 🔍 Checklist

---

## 📝 Additional Notes

The immediate trigger for the 413 is the default `apiMaxLength` of `100mb` defined in `server/api/index.js`. To import a project larger than 100 MB without changing source code, add the following line to `server/_appdata/settings.js`:

```js
apiMaxLength: '256mb',
```

Browser behaviour for large uploads is not standardised. Chrome typically receives the 413 response cleanly (the `status === 413` branch fires and shows the server message verbatim). Firefox and some other clients abort the TCP connection before receiving the response (the `status === 0` branch fires and shows the configured limit as a hint). Both paths now produce an actionable message instead of the generic fallback.

The server log will now show:

```log
[ERR] api request rejected: The submitted content exceeds the maximum allowed size (100mb)
```

The final result on the ui is:
error_max_size

### DIFF
--- a/client/src/app/_services/project.service.ts
+++ b/client/src/app/_services/project.service.ts
@@ -145,14 +145,7 @@ export class ProjectService {
             }
             subject.next(true);
         }, err => {
-            console.error(err);
-            var msg = '';
-            this.translateService.get('msg.project-save-error').subscribe((txt: string) => { msg = txt; });
-            this.toastr.error(msg, '', {
-                timeOut: 3000,
-                closeButton: true,
-                disableTimeOut: true
-            });
+            this.notifySaveError(err);
             subject.next(false);
         });
         return subject;
@@ -933,6 +926,8 @@ export class ProjectService {
             msg = this.translateService.instant('msg.project-save-unauthorized');
         } else if (err.status === 413) {
             msg = err.error?.message || err.message || err.statusText;
+        } else if (err.status === 0 && this.serverSettings?.apiMaxLength) {
+            msg = `Upload failed: project size may exceed the server limit (${this.serverSettings.apiMaxLength})`;
         }
         if (msg) {
             this.toastr.error(msg, '', {
@@ -1332,6 +1327,7 @@ export class ProjectService {
 export class ServerSettings {
     version: string;
     secureEnabled: boolean;
+    apiMaxLength?: string;
 }
 
 export enum SaveMode {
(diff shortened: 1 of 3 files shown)


unocelli (changes requested):
Hi @zarch Thanks for the PR. About theme.scss don't have to be change, I think you have to reinstall node_modules and build the client again

PR #2246, opened 2026-03-11
Fix/view items auto cleanup

zarch:
## 📌 Description

FUXA project files accumulate orphaned entries in `view.items` — the dictionary that maps SVG element IDs to their gauge configuration. An orphan is an entry whose key no longer appears as an `id="…"` attribute anywhere in `view.svgcontent`.

Orphans are created whenever an SVG element is removed from the canvas (Delete key, Cut, Undo of an add) because the existing `onGaugeRemoved` callback registered in `editor.component.ts` is never actually fired by the SVG editor library. Over time, across many edit-save cycles, these dead entries inflate the project JSON file.

The method `cleanView()` that removes orphans was already implemented and correct, but was only called in two manual places (the Play button and the "Clean View" menu item), not on the automatic save/load path.

### What was changed?

Two call sites added — one per file, each independently cherry-pickable:

1. **`client/src/app/_services/project.service.ts`**
   `cleanView()` is now called at the top of `setView()` and `setViewAsync()`, immediately before the view is serialised and sent to the server. Orphans accumulated during an editing session are stripped before they can be persisted.

2. **`client/src/app/editor/editor.component.ts`**
   `cleanView()` is now called in `loadView()` before the `setTimeout` loop that initialises gauge components from `view.items`. Orphans stored in a previously saved file are removed before any Angular component is created for them.

### Why is this needed?

The fix stops the bloat accumulation loop:

```
Before fix:
  Edit (add + undo) → orphan in view.items → save → orphan persisted → reload → orphan
  re-initialised → save again → grows indefinitely

After fix:
  Edit (add + undo) → orphan in view.items → save → cleanView() strips it → only live
  items are persisted → reload → cleanView() strips any survivors → stable
```

No new logic is introduced. The fix re-uses the battle-tested `cleanView()` method already in the repository.

---

## 🧪 Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Other

---

## 🚫 Build Artifacts Check

Please confirm:

---

## 🔍 Checklist

---

### Each commit is independently useful

The two commits are intentionally separate so that a maintainer can cherry-pick either half:

| Commit | Benefit |
|---|---|
| `fix(project-service)` | Stops orphans from being written to disk on every save |
| `fix(editor)` | Stops orphans already in a saved file from being loaded into memory |

The save-path commit has higher impact in practice; the load-path commit is complementary.

### Risk

Low. `cleanView()` is a pure read-then-delete operation on an in-memory dict. It is already executed on the Play path, so it has been in production use. The regex that parses SVG ids is unchanged.

### Performance

`cleanView()` runs a single regex pass over `view.svgcontent` and one `Object.keys()` iteration. For a view with a large SVG, this is a one-time cost at save/load, not in the render loop. No measurable impact is expected.

### DIFF
--- a/client/src/app/editor/editor.component.ts
+++ b/client/src/app/editor/editor.component.ts
@@ -526,6 +526,7 @@ export class EditorComponent implements OnInit, AfterViewInit, OnDestroy {
 
                 // check gauge to init
                 this.gaugesRef = {};
+                if (v) { this.projectService.cleanView(v); }
                 setTimeout(() => {
                     for (let key in v.items) {
                         let ga: GaugeSettings = this.getGaugeSettings(v.items[key]);


unocelli:
Ciao @zarch, Ottimo lavoro, Grazie!

PR #2200, opened 2026-02-13
feat(modbus): implement COM port sharing and robust connection management (#211)

lookmydog:
## Description
This PR refactors the Modbus communication layer to support multiple slave devices sharing a single COM port. It introduces a centralized management system to ensure stability and efficiency in industrial environments.

## Key Mechanisms Implemented:
1. **ConnectionManager**: A centralized controller to manage port states and lifecycle.
2. **Reference Counting**: Tracks the number of active devices per port to safely manage opening/closing.
3. **Queue System**: Serializes concurrent requests to prevent data collisions on the physical layer.
4. **Skip Mechanism**: Automatically skips unreachable or faulty devices to maintain overall system performance.
5. **Timeout Handling**: Improved detection and recovery for unresponsive communication sessions.
6. **Concurrency Control**: Robust logic to handle simultaneous connection attempts across multiple slaves.
7. **State-Sync Mechanism**: Maintains a clear separation between desired state (intended value) and actual state (device feedback), and continuously reconciles the difference to ensure reliable and consistent communication with the underlying devices. (referencing the approach from @alext-extracellular).

## Impact
- Allows more flexible hardware configurations (RS-485 daisy-chaining).
- Reduces resource overhead by reusing port instances.
- Increases communication reliability under high load.

## Related Issue
Fixes #211

## ✅ Checklist
- [x] Implement core mechanisms (1 to 6)
- [x] Fix connection drop issue during normal operations (e.g., adding tags)
- [x] Implement `state-sync` mechanism
- [x] Add `connect-skipped` status and UI visual indicator
- [x] Update all i18n translation files for the new status
- [x] Replace `console.log` with FUXA built-in logger
- [x] Perform testing on physical devices or simulators
  - [x] **Test 01 (RTU Shared Port):** Two slave devices sharing a single COM port; verify normal polling operations without data collisions or conflicts.
  - [x] **Test 02 (TCP ReuseSerial):** Verify connection and queueing stability under TCP shared socket mode (e.g., connecting via an RS-485 to Ethernet Gateway).
  - [x] **Test 03 (Standard TCP):** Verify connection and data operations (read/write) under standard Modbus TCP mode (independent IP/Socket) to ensure legacy logic remains unaffected.
  - [x] **Test 04 (State-Sync):** Test adding tags or frequently writing values (SetValue) during runtime; verify that desired and actual states synchronize correctly without causing connection drops.
[...]

PR #2262, opened 2026-03-15
WIP: adjust modbus sync logic and implement connection manager

alext-extracellular:
## 📌 Description

A similar PR to this one: https://github.com/frangoteam/FUXA/pull/2200
However I have implemented a different approach for synchronising with modbus devices. Instead of using a queue or sending the commands whenever the UI is used, the UI only modifies a local version of the state, which is then synchronised to the remote devices on the configured polling interval. This reduces the chance of timeout errors, and also means that a delay cannot build up between FUXA and the device on the other end of the queue. In this case, the delay between an action in FUXA and an update on the remote device will always be equal to the polling interval. It allows for setting higher polling frequencies with less chance of errors, as there are no user input triggered commands that can get sent at the same time and overload the bus.

To support multiple server devices on one RTU or TCP device, a connection manager has been added, which will create a key and shared connection that any other devices will reuse. this is the default behaviour now when no "Socket Reuse" option is selected.

I will mark as WIP for now, as I want to run some more tests in the coming week on my own hardware, but initial tests have succeeded and provided better performance.

---

## 🧪 Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Other

---

## 🚫 Build Artifacts Check

Please confirm:

---

## 🔍 Checklist

---


alext-extracellular:
tabs have been replaced by spaces

PR #2336, opened 2026-04-29
Fix/opcua publishing interval

bartei:
## 📌 Description

The official image doesn't honor the refresh rate configured for the opc-ua subscriptions.

This PR uses the already present variable to handle the refresh rate as requested.
---

## 🧪 Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Other

---

## 🚫 Build Artifacts Check

Please confirm:

- [X] I did NOT commit `/client/dist`
- [X] I did NOT commit generated Angular build output
- [X] Only source files are included

---

## 🔍 Checklist

- [X] Code follows project coding standards
- [X] I tested my changes locally
- [X] Documentation updated if required

---

- [ ] All links were tested

---

## 📝 Additional Notes


unocelli:
Hi @bartei Thanks for the PR. The OPC UA change makes sense in principle: using `data.polling` for `requestedPublishingInterval` aligns the subscription publishing rate with the user-configured device update rate.

I would request a couple of changes before merging:

1. Please keep the default behavior at `500 ms` when `data.polling` is not configured. Changing the fallback from `500` to `100` makes every existing/default OPC UA setup faster by default, which may increase traffic and CPU load on installations with many tags or slower OPC UA servers.

Suggested shape:

```js
const polling = Number(data.polling) || 500;

requestedPublishingInterval: polling
```

2. Please remove the Dockerfile and GHCR workflow changes from this PR.

bartei:
All done and ready for your review

unocelli:
Hi @bartei Thanks a lot!

PR #2355, opened 2026-05-07
The Fix for Issue #1865 — "Negative Time in Min-Max in Alarm Configuration"

### LINKED ISSUE #1865: [BUG] Negative Time in Min-Max in Alarm Configuration
Inside alarms configuration panel it's possible to set an negative value inside the Time in Min-Max range field.
I don't know if there is any consequences inside the backend but this shouldn't be possible to do that.

Image

prashantmavani0706:
What problem does this solve?
Fixes #1865 — Negative time delay in alarm Min-Max configuration
Problem: The alarm configuration panel allowed users to type negative values into the Time Delay field. A negative time delay causes alarms to fire immediately, bypassing the intended delay behavior.

Approach (based on owner feedback):
Rather than changing default behavior in shared code, NumberOnlyDirective now exposes two opt-in inputs with backward-compatible defaults:

[allowNegative] — defaults to true (preserves existing behavior for all current consumers)
[allowDecimal] — defaults to true (preserves existing behavior for all current consumers)

Only the 5 timedelay fields in alarm-property.component.html explicitly opt into strict mode:
html<input numberOnly [allowNegative]="false" [allowDecimal]="false" ...>
Files changed: 2

number.directive.ts — adds configurable @Input() options
alarm-property.component.html — timedelay fields opt into strict mode

Zero regressions — all other consumers of numberOnly across the codebase are unaffected.

Type of Change: Bug fix


unocelli:
Hi @prashantmavani0706 Thanks for the PR but it currently does not compile, please check it

prashantmavani0706:
@unocelli -Thank you for the review! The previous commit accidentally wrapped NumberOrNullOnlyDirective in a block comment, causing the compile error. This is now fixed — NumberOrNullOnlyDirective is restored unchanged, and only NumberOnlyDirective was modified to add the regex validation and preventDefault() call that blocks the minus key. The range-number.component.html change remains the same.

unocelli:
Hi @prashantmavani0706 Thanks for the PR. The use case is valid, but I think we should avoid changing default behavior in shared code.

This PR changes both `numberOnly` and `range-number`, and both are reused in several places across FUXA. Changing their default behavior can introduce regressions in consumers that currently rely on the existing numeric handling.

A safer approach would be to make the behavior configurable while keeping defaults backward-compatible. For example, `numberOnly` could expose options like `[allowNegative]` and `[allowDecimal]`, defaulting to the current permissive behavior. Then fields that require only positive integers can explicitly use:

`<input numberOnly [allowNegative]="false" [allowDecimal]="false">`

Similarly, `range-number` should expose/pass through explicit options instead of changing behavior for every usage. Consumers that need signed decimal ranges can opt in, while existing usages keep their current behavior.

prashantmavani0706:
Hi @unocelli  — Thanks for the guidance. Updated based on your feedback. Here's what changed:
NumberOnlyDirective now has two opt-in @Input() properties — allowNegative and allowDecimal — both defaulting to true to exactly match the original permissive behaviour. All existing usages across the codebase are untouched.
alarm-property.component.html — only the 5 timedelay fields explicitly opt into strict mode with [allowNegative]="false" [allowDecimal]="false", exactly as you suggested.
Only 2 files changed. Zero regressions — all existing consumers keep their current behaviour since both inputs default to true.

unocelli:
Hi @prashantmavani0706 Thanks for the update. I tested it again and I think there is still a regression risk.

With this change, some existing shape/control properties can no longer be configured with negative values, but those negative values are valid in the editor. So the fix affects more than the alarm time delay case.
image


I think it would be safer not to change the behavior of the shared numeric/range input globally. A better approach may be to introduce a dedicated range-number variant or a configurable option for this specific use case, so alarm time delay can reject negative values without breaking fields that legitimately support negatives elsewhere.

prashantmavani0706:
Hi @unocelli  — Thank you for the detailed feedback and the screenshot. I understand the concern now.
Before updating the PR, I wanted to confirm the approach with you first.
My proposed solution:
Add a new dedicated directive nonNegativeIntegerOnly to number.directive.ts — blocks "-", ".", "e" and non-digit keys. numberOnly and numberOrNullOnly are not touched at all.
Register NonNegativeIntegerOnlyDirective in app.module.ts.
Replace numberOnly with nonNegativeIntegerOnly on the 5 timedelay fields in alarm-property.component.html only.
This means numberOnly, range-number, and all other shared components remain completely unchanged — zero risk to shape settings, slider, chart, or any field that legitimately supports negative values.
Could please check if this align with what you had in mind? Please let me know if you have any other suggestion. Thanks.

unocelli:
Hi @prashantmavani0706 Yes, this approach aligns with what I had in mind.

Keeping `numberOnly`, `numberOrNullOnly`, `range-number`, and the shared shape/control inputs unchanged avoids the regression risk. Applying a dedicated `nonNegativeIntegerOnly` directive only to the alarm `timedelay` fields sounds like a safer and more focused fix for this issue.

Please also make sure paste/input cases are handled, not only keydown, so negative values cannot still be inserted via paste or browser number input behavior. Adding `min="0"` and `step="1"` on those fields would also be good.

prashantmavani0706:
Hi @unocelli  — Thank you for the approval and the additional requirements. Before updating the PR, I wanted to confirm the complete approach with you:
For the nonNegativeIntegerOnly directive, I plan to handle all three entry vectors you mentioned:
- keydown — blocks -, ., e, E, + and all non-digit keys
- paste — intercepts paste events and strips any non-digit characters so negative or decimal values cannot be inserted via clipboard
- input — clamps the value to a non-negative integer after browser number spinner changes (e.g. clicking the down arrow past 0)

For the timedelay fields in alarm-property.component.html, I will also add min="0" and step="1" attributes to all 5 fields as you suggested.

This means the total changes are 4 files only:
1)number.directive.ts — new NonNegativeIntegerOnlyDirective appended at the end; numberOnly and numberOrNullOnly are completely untouched
2) number.directive.spec.ts — unit tests covering all three entry vectors (keydown, paste, input) plus min/step attribute cases and regression tests confirming numberOnly original behaviour is preserved
3) app.module.ts — import and declaration of the new directive only
4) alarm-property.component.html — exactly 5 timedelay fields updated with nonNegativeIntegerOnly, min="0" and step="1"
numberOnly, numberOrNullOnly, range-number, and all other shared components remain completely untouched.

Does this fully address what you had in mind? Happy to proceed once you confirm. Thanks.

prashantmavani0706:
> Hi @unocelli — Thank you for the approval and the additional requirements. Before updating the PR, I wanted to confirm the complete approach with you: For the nonNegativeIntegerOnly directive, I plan to handle all three entry vectors you mentioned:
> 
> * keydown — blocks -, ., e, E, + and all non-digit keys
> * paste — intercepts paste events and strips any non-digit characters so negative or decimal values cannot be inserted via clipboard
> * input — clamps the value to a non-negative integer after browser number spinner changes (e.g. clicking the down arrow past 0)
> 
> For the timedelay fields in alarm-property.component.html, I will also add min="0" and step="1" attributes to all 5 fields as you suggested.
> 
> This means the total changes are 4 files only: 1)number.directive.ts — new NonNegativeIntegerOnlyDirective appended at the end; numberOnly and numberOrNullOnly are completely untouched 2) number.directive.spec.ts — unit tests covering all three entry vectors (keydown, paste, input) plus min/step attribute cases and regression tests confirming numberOnly original behaviour is preserved 3) app.module.ts — import and declaration of the new directive only 4) alarm-property.component.html — exactly 5 timedelay fields updated with nonNegativeIntegerOnly, min="0" and step="1" numberOnly, numberOrNullOnly, range-number, and all other shared components remain completely untouched.
> 
> Does this fully address what you had in mind? Happy to proceed once you confirm. Thanks.

Hi @unocelli - could you please help to provide your feedback on above?

unocelli:
Hi @prashantmavani0706, yes, this approach looks good to me.

Keeping `numberOnly`, `numberOrNullOnly`, `range-number`, and all shared inputs unchanged is the right direction and keeps the fix scoped only to the alarm `timedelay` fields.

The only small suggestion I have is about paste handling: instead of stripping non-digit characters, I think it would be safer to reject invalid pasted values entirely. For example, pasting `-5` should not silently become `5`, because that changes the user input.

So the expected behavior would be:
- allow only non-negative integers;
- block `-`, `.`, `e`, `E`, `+` and non-digit keys;
- reject invalid paste values;
- sanitize/clamp invalid input values from browser controls if needed;
- add `min="0"` and `step="1"` to the 5 `timedelay` fields.

The proposed test coverage also sounds good, especially the regression test confirming that `numberOnly` keeps its existing behavior.

prashantmavani0706:
Hi @unocelli — Thanks again for suggestion. Updated the PR based on your latest feedback. Here is what is now in place:

nonNegativeIntegerOnly directive handles all three entry vectors:
1. keydown —  blocks -, ., e, E, + and all non-digit keys
2. paste — rejects invalid pasted values entirely; pasting -5 is discarded as a whole and does not silently become 5
3. input — clamps any value from the browser number spinner to a non-negative integer (e.g. spinning below 0 resets to 0)

alarm-property.component.html — all 5 timedelay fields now have min="0" and step="1".

4 files changed in total:
1. number.directive.ts — new NonNegativeIntegerOnlyDirective appended; numberOnly and numberOrNullOnly are byte-for-byte identical to the original
2. number.directive.spec.ts — 27 unit tests covering keydown, paste, input, min/step attributes, and regression tests confirming numberOnly keeps its original behaviour
3. app.module.ts — import and declaration of the new directive
4. alarm-property.component.html — 5 timedelay fields updated

numberOnly, numberOrNullOnly, range-number, and all other shared components are completely untouched. 

Could you please help to check and let me know in case of any correction needed.

unocelli:
Hi @prashantmavani0706, thanks for the update. The scoped `nonNegativeIntegerOnly` approach looks good, but the latest branch still has a few issues.

I tested it locally and the build currently fails because `app.module.ts` imports/declarations include unrelated components that are not present in the branch. Please keep only the `NonNegativeIntegerOnlyDirective` registration there.

Also, `range-number.component.html` is still modified in the PR, but this file should remain unchanged for this fix.

One small implementation note: after a valid paste, please make sure Angular/ngModel is updated correctly, since the directive writes directly to `input.value`.

Once these are cleaned up, the fix looks aligned with the intended approach.

unocelli:
Hi @prashantmavani0706 Good job, Thanks a lot

prashantmavani0706:
Hi @unocelli - Thanks to you as well for your guidance and support.

unocelli:
Hi @prashantmavani0706, thanks again for the PR.

After merging, I made a small follow-up fix: Ctrl/Cmd shortcuts are now allowed, and valid paste values are handled by the browser normally instead of being inserted manually by the directive. This fixes a double-paste issue while still rejecting invalid pasted values like `-5`, decimals, or text.

I also removed the frontend unit test file for now, since we are currently not maintaining frontend unit tests and are evaluating a different approach for frontend testing.

PR #2364, opened 2026-05-14
fix(docker): add missing ODBC runtime

### LINKED ISSUE #2282: [BUG] ODBC MariaDB driver fails to load in Docker image - missing libmariadb.so.3
**Describe the bug**

The MariaDB ODBC driver (`libmaodbc.so`) in the Docker image (`frangoteam/fuxa:latest`) cannot be loaded because its runtime dependency `libmariadb.so.3` is missing from the final stage (slim) image.

This causes all ODBC connections using `DRIVER=MariaDB` to fail with:

```
[ERR] 'test' try to connect error! Error: [odbc] Error connecting to the database
```

**To Reproduce**

1. Run FUXA using Docker:
```yml
services:
  fuxa:
    image: frangoteam/fuxa:latest
    restart: unless-stopped
    volumes:
      - './appdata:/usr/src/app/FUXA/server/_appdata'
      - './db:/usr/src/app/FUXA/server/_db'
      - './logs:/usr/src/app/FUXA/server/_logs'
      - './images:/usr/src/app/FUXA/server/_images'
    ports:
      - '1881:1881'
```
2. Configure an ODBC connection in the UI with `DRIVER=MariaDB`
3. Connection fails repeatedly

**Expected behavior**
The ODBC connection using `DRIVER=MariaDB` should work out of the box in the Docker image without needing to manually install additional libraries.

**Screenshots**
none

**Example project**
none

**Environment**
look my docker compose 

**Cause**

In the Dockerfile, the `odbc-mariadb` package is installed in the **server-builder stage**, which automatically pulls in `libmariadb3` as a dependency. However, in the **final runner stage** (`node:18-bookworm-slim`), only `unixodbc` is installed — `libmariadb3` is not included.

dev-soragoto:
## 📌 Description

Install odbc-postgresql, libsqliteodbc, tdsodbc and odbc-mariadb in the runner stage and collect their drivers into /usr/lib/odbc/. Copy MySQL Connector/ODBC and MSSQL drivers from the builder stage since they are not available in Debian repos.

Fixes #2282

---

## 🧪 Type of Change

- [x] Bug fix

---

## 🚫 Build Artifacts Check

---

## 🔍 Checklist

---

- [ ] All links were tested

---

## 📝 Additional Notes

Tested on x64 via windows11. Verified all 9 drivers registered in odbcinst.ini are present in the final image via `odbcinst -q -d`.

```pwsh
PS C:\Users\soragoto\FUXA> docker run --rm fuxa-test sh -c 'odbcinst -q -d'
[MySQL]
[MySQL ANSI]
[PostgreSQL]
[PostgreSQL ANSI]
[SQLite]
[SQLite3]
[FreeTDS]
[MSSQL]
[MariaDB]
PS C:\Users\soragoto\FUXA> docker run --rm fuxa-test sh -c 'ls -la /usr/lib/odbc'
total 50004
drwxr-xr-x 1 root root     4096 May 25 12:30 .
drwxr-xr-x 1 root root     4096 May 25 12:29 ..
-rw-r--r-- 1 root root   336208 May 25 12:29 libmaodbc.so
-rwxr-xr-x 1 root root 24524112 May 25 12:30 libmyodbc8a.so
-rwxr-xr-x 1 root root 24543144 May 25 12:30 libmyodbc8w.so
-rw-r--r-- 1 root root   198128 May 25 12:29 libsqlite3odbc-0.9998.so
lrwxrwxrwx 1 root root       24 May 25 12:29 libsqlite3odbc.so -> libsqlite3odbc-0.9998.so
-rw-r--r-- 1 root root   470360 May 25 12:29 libtdsodbc.so
drwxr-xr-x 2 root root     4096 May 25 12:30 private
-rw-r--r-- 1 root root   535600 May 25 12:29 psqlodbca.so
-rw-r--r-- 1 root root   572512 May 25 12:29 psqlodbcw.so
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,7 +37,8 @@ WORKDIR /usr/src/app/FUXA/odbc
 COPY odbc/ ./
 RUN if [ "$INSTALL_ODBC" = "true" ]; then \
     dos2unix install_odbc_drivers.sh && chmod +x install_odbc_drivers.sh && ./install_odbc_drivers.sh; \
-    fi
+    fi \
+    && mkdir -p /usr/lib/odbc /opt/microsoft
 
 # 3. Copy server source, build, then cleanup
 WORKDIR /usr/src/app/FUXA/server
@@ -51,10 +52,19 @@ ARG INSTALL_ODBC=true
 WORKDIR /usr/src/app/FUXA
 
 # Install ONLY runtime libraries
-RUN apt-get update && apt-get install -y \
-    sqlite3 libsqlite3-0 \
-    $( [ "$INSTALL_ODBC" = "true" ] && echo "unixodbc" ) \
+RUN apt-get update \
+    && apt-get install -y \
+        sqlite3 libsqlite3-0 \
+        $( [ "$INSTALL_ODBC" = "true" ] && echo "unixodbc odbc-mariadb odbc-postgresql libsqliteodbc tdsodbc" ) \
+    && if [ "$INSTALL_ODBC" = "true" ]; then \
+        mkdir -p /usr/lib/odbc && \
+        find /usr/lib -path '*/odbc/*.so' -exec cp {} /usr/lib/odbc/ \; ; \
+    fi \
     && rm -rf /var/lib/apt/lists/*
+
+# Copy MySQL and MSSQL ODBC drivers from builder (not available in Debian repos)
+COPY --from=server-builder /usr/lib/odbc/ /usr/lib/odbc/
+COPY --from=server-builder /opt/microsoft/ /opt/microsoft/
 
 # 1. Copy Server
 COPY --from=server-builder /usr/src/app/FUXA/server ./server


unocelli:
Hi @dev-soragoto Thanks, this looks like the right fix for the MariaDB runtime issue.

One small thing: what do you think about making the copy step a bit less dependent on the exact Debian multiarch path?

In the runner stage, this line currently assumes the path can be built from `uname -m`:

cp /usr/lib/$(uname -m)-linux-gnu/odbc/libmaodbc.so /usr/lib/odbc/libmaodbc.so; \

Would you be open to replacing it with something like this?

maodbc_path="$(find /usr/lib -path '*/odbc/libmaodbc.so' -print -quit)" && \
test -n "$maodbc_path" && \
cp "$maodbc_path" /usr/lib/odbc/libmaodbc.so; \

That should keep the current x86_64 and ARM64 cases working, while avoiding a hardcoded assumption about the architecture directory name.

If you agree, I think this would be good to merge after that change.

dev-soragoto:
@unocelli Thanks for the suggestion! I've updated the PR:

- Replaced the `uname -m` hardcoded path with `find` as you suggested, so it works across architectures without assumptions about the directory name
- Updated the PR description to reflect the actual changes
- Also noticed that other drivers registered in `odbcinst.ini` (PostgreSQL, SQLite, FreeTDS, MySQL, MSSQL) were missing from the runner stage entirely, so I've fixed those as well

unocelli:
Hi @dev-soragoto Good job, Thanks a lot!

PR #2375, opened 2026-05-22
Add Omron NJ/NX EtherNet/IP device driver

Pasqua11:
Adds a new OmronEthernetIP device type for Omron NJ/NX series PLCs. These controllers use CIP symbolic tag messaging, a protocol the existing EthernetIP (PCCC/Allen-Bradley) driver cannot speak.

The driver wraps the optional 'omron-ethernet-ip' npm package and is registered as an optional plugin (like nodepccc/EthernetIP), so it is not a bundled dependency: if the package is not installed the device type is simply unavailable and FUXA still starts normally.

Server:
- runtime/devices/omron-ethernetip/index.js - driver wrapper
- runtime/devices/device.js - register the OmronEthernetIP device type
- runtime/plugins/index.js - register omron-ethernet-ip as a plugin

Client:
- DeviceType and PluginType enums
- device property editor: host:port address field
- tag editing and plugin icon reuse the EthernetIP handlers

## 📌 Description

Please describe the changes introduced by this Pull Request.

- What problem does this solve?
- What was changed?
- Why is this needed?

For documentation-only PRs, focus on the Documentation checklist.

---

## 🧪 Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Other

---

## 🚫 Build Artifacts Check

Please confirm:

- [ ] I did NOT commit `/client/dist`
- [ ] I did NOT commit generated Angular build output
- [ ] Only source files are included

---

## 🔍 Checklist

- [ ] Code follows project coding standards
- [ ] I tested my changes locally

---

- [ ] All links were tested

---

## 📝 Additional Notes

### DIFF
--- a/client/src/app/_models/device.ts
+++ b/client/src/app/_models/device.ts
@@ -246,6 +246,7 @@ export enum DeviceType {
     MQTTclient = 'MQTTclient',
     internal = 'internal',
     EthernetIP = 'EthernetIP',
+    OmronEthernetIP = 'OmronEthernetIP',
     ODBC = 'ODBC',
     ADSclient = 'ADSclient',
     GPIO = 'GPIO',
--- a/client/src/app/_models/plugin.ts
+++ b/client/src/app/_models/plugin.ts
@@ -17,6 +17,7 @@ export enum PluginType {
     Raspberry = 'Raspberry',
     SiemensS7 = 'SiemensS7',
     EthernetIP = 'EthernetIP',
+    OmronEthernetIP = 'OmronEthernetIP',
     MELSEC = 'MELSEC',
     REDIS = 'REDIS'
 }
--- a/client/src/app/device/device-list/device-list.component.ts
+++ b/client/src/app/device/device-list/device-list.component.ts
@@ -274,7 +274,7 @@ export class DeviceListComponent implements OnInit, AfterViewInit {
 
     isToEdit(type, tag: Tag) {
         if (type === DeviceType.SiemensS7 || type === DeviceType.ModbusTCP || type === DeviceType.ModbusRTU ||
-            type === DeviceType.internal || type === DeviceType.EthernetIP || type === DeviceType.FuxaServer ||
+            type === DeviceType.internal || type === DeviceType.EthernetIP || type === DeviceType.OmronEthernetIP || type === DeviceType.FuxaServer ||
             type === DeviceType.OPCUA || type === DeviceType.GPIO || type === DeviceType.ADSclient ||
             type === DeviceType.WebCam || type === DeviceType.MELSEC || type === DeviceType.REDIS) {
             return true;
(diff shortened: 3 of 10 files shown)


unocelli:
Hi @Pasqua11 Thanks for the great work on this driver,

I pushed one more commit to the PR branch:

52bcf008 Use overload guard for Omron polling recovery

A quick note on what happened here: my first attempt was a bit too defensive. It added a timeout wrapper around every readTag() call, but after looking at it again that felt heavier than needed for the normal polling path.

The latest commit keeps the read path simpler: successful polling still just does await client.readTag(...). The recovery is handled through the existing working/overload guard instead.

The main reasons for the change are:

- if all reads fail, we should not refresh lastReadTimestamp, otherwise FUXA may treat stale values as fresh;
- if polling gets stuck, after repeated overloads the driver closes the Omron client and emits connect-error, so the normal reconnect flow can recover;
- there is also a small connection version guard to avoid an old in-flight read updating values after a reconnect.

I added mocked tests for the wrapper behaviour and ran:

- node --check server/runtime/devices/omron-ethernetip/index.js
- node --check server/test/devices/omronEthernetip.test.js
- npm test -- test/devices/omronEthernetip.test.js -> 4 passing

I cannot test this with a real Omron NJ/NX PLC from my side. Are you OK with this wrapper behaviour? If yes, could you please test the updated branch on real hardware, especially read/write of scalar tags, BOOL writes with 0.0.2, STRING tags if available, and disconnect/reconnect behaviour?

If something does not match the Omron client behaviour in practice, we can adjust it before merge.

Pasqua11:
Eccolo, pronto da copiare e incollare nel commento di risposta a `unocelli`:

```
Hi @unocelli, thanks for the update — I tested the updated branch (52bcf00) on a real Omron NX PLC. The overall behaviour looks good. 👍

What I tested:
- BOOL writes (0.0.2): working correctly, both with button and switch widgets — the bit toggles reliably on the PLC.
- Scalar read: OK (INT/REAL values read correctly).
- Disconnect / reconnect: works as designed. When I unplugged the PLC, after a few overloads the driver logged `polling blocked, reconnecting`, closed the client and emitted `connect-error`. It then kept retrying every 5s without getting stuck (left it ~12 min), and reconnected automatically as soon as the PLC came back (`restored 0/0 values`, values resumed). Importantly it did not refresh stale values with a fresh timestamp.

Two minor (cosmetic) notes — not blockers:
1. When the client is set to `null` mid-cycle during recovery, the still-running polling loop logs one `Cannot read properties of null (reading 'readTag')` error per tag (~200 lines in my case). Functionally harmless, but very noisy — maybe break out of the loop once the connection version changes / client is null.
2. The reconnect attempts log `connect error! undefined (n)` — the error detail shows `undefined` instead of a meaningful message.

```

unocelli:
Hi @Pasqua11 Thanks a lot! Good job!

PR #2388, opened 2026-06-01
Update fuxa-ack-alarm.html

Nerzh29:
## 📌 Description

Please describe the changes introduced by this Pull Request.

- **What problem does this solve?**
  The `ack-alarm` Node-RED node requires users to select an alarm from a static dropdown menu because the `alarmName` field is marked as `required: true`. This strictly blocks the Node-RED editor from deploying flows unless a static alarm is chosen, which directly contradicts the node's built-in help documentation stating that users can pass dynamic overrides via `msg.alarmName` or `msg.payload`.

- **What was changed?**
  In `fuxa-ack-alarm.html`: Changed the `alarmName` property validation from `required: true` to `required: false`.

- **Why is this needed?**
  This change is needed to allow a fully dynamic, message-driven workflow without being forced by the Node-RED editor UI to select a placeholder static alarm just to be able to deploy the flow.

---

## 🧪 Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Other

---

## 🚫 Build Artifacts Check

Please confirm:

---

## 🔍 Checklist

---

- [ ] All links were tested

---

## 📝 Additional Notes

By changing `required` to `false`, the editor validation passes even if the field is empty, allowing the node to gracefully rely on runtime incoming messages (`msg.alarmName`).

### DIFF
--- a/node-red/node-red-contrib-fuxa/nodes/fuxa-ack-alarm.html
+++ b/node-red/node-red-contrib-fuxa/nodes/fuxa-ack-alarm.html
@@ -4,7 +4,7 @@
         color: '#a6bbcf',
         defaults: {
             name: {value:""},
-            alarmName: {value:"", required:true},
+            alarmName: {value:"", required:false},
             types: {value:""}
         },
         inputs:1,


unocelli:
Hi @Nerzh29 Thanks for the contribution. The proposed behavior makes sense. However, this PR appears to update the legacy node-red/node-red-contrib-fuxa copy. The active FUXA Node-RED integration seems to load its nodes from server/integrations/node-red/node-red-contrib-fuxa, where alarmName would remain required.

Could you please confirm whether the current file location is intentional? If not, the change should probably be applied to the active copy instead.

unocelli:
Hi @Nerzh29 Thanks for the PR!

PR #2410, opened 2026-06-22
fix: gauge import fallback in electron

VeikkoAJ:
## 📌 Description

**Bug:**

Gauges are not rendered in electron app.  The gauges are shown as white boxes. Gauges work otherwise normally. 

Example screenshots below from latest electron build. Left side is electron app and right side is browser:
fuxa gauge bug

Same happens in editor:
fuxa gauge bug editor

**Reason for the bug:** 

In `gauge.js:974` else if check
```js
} else if (typeof module !== 'undefined' && (module.exports != null)) {
```
should evaluate false but evaluates true because `nodeIntegration: true` injects Node.js globals (in this case `module`) which in electron app causes `gauge.js` to use headless version exports. 

**Solution:** 
Added fallback require to `/ngx-gauge.component.ts`  that imports the gauge for electron app.  

Alternative fix would be to check if window is defined in `gauge.js:974`
```js
} else if (typeof module !== 'undefined' && module.exports != null && typeof window === 'undefined') {
```

After fix:
fuxa gauge bug fixed

And editor: 
fuxa gauge bug editor fixed

I tested the fix on Windows. Any Feedback on the preferred fix approach welcome.

## 🧪 Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Other

---

## 🚫 Build Artifacts Check

Please confirm:

---

## 🔍 Checklist

## 📝 Additional Notes

### DIFF
--- a/client/src/app/gui-helpers/ngx-gauge/ngx-gauge.component.ts
+++ b/client/src/app/gui-helpers/ngx-gauge/ngx-gauge.component.ts
@@ -7,6 +7,15 @@ import { Subject, interval, take, takeUntil } from 'rxjs';
 declare const Gauge: any;
 declare const Donut: any;
 
+// In Electron (nodeIntegration: true), gauge.js detects module.exports and skips setting window globals.
+// Fall back to require() to get the CommonJS export directly.
+function getGaugeClasses(): { Gauge: any, Donut: any } {
+    if (typeof Gauge !== 'undefined') {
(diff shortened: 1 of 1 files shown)


unocelli:
Hi @VeikkoAJ Good job! Thanks a lot!

PR #2412, opened 2026-06-24
fix: re-trigger resize after async view load in grid cards

VeikkoAJ:
## 📌 Description

**Problem:** 
Grid view cells clip instead of properly scaling, regardless of scale mode. The bug is caused by the async loading of the grid view. Gridster fires `itemResizeCallback` before item content is set. 

**How to replicate:** 
1. Create a view with text like in the screnshot below: fuxa grid view example 1
2. Add the view to grid and scale the cell smaller than the actual size of the view: fuxa grid view example 2
3. The view is clipped: fuxa grid view broken

**Fix:**
 Call `this.gridOptions.itemResizeCallback` at the end of `initCallback`, after the view is loaded and change detection has run. 

What it looks like in after fix: 
fuxa grid view fixed

---

## 🧪 Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Other

---

## 🚫 Build Artifacts Check

Please confirm:

---

## 🔍 Checklist

### DIFF
--- a/client/src/app/cards-view/cards-view.component.ts
+++ b/client/src/app/cards-view/cards-view.component.ts
@@ -119,6 +119,7 @@ export class CardsViewComponent implements OnInit, AfterViewInit {
                 }
                 this.itemChange(item, itemComponent);
                 this.changeDetector.detectChanges();
+                this.gridOptions.itemResizeCallback?.(item, itemComponent)
             }
         };
         this.dashboard.push(item);


unocelli:
Hi @VeikkoAJ Good job! Thanks a lot!

VeikkoAJ:
> Hi @VeikkoAJ Good job! Thanks a lot!

No problem! Can you run the headless build and electron build actions. The latest version in both of those is 1.3.1

unocelli:
Yes, I'm trying to fix it

PR #2437, opened 2026-07-14
fix(s7): resolve Linux segfaults and stability issues in S7 driver

JordiB-Inserma:
## Summary

- Fixes segfaults on Linux caused by missing \
ew\ keyword in S7Client instantiation combined with destructor bugs in node-snap7
- Fixes permanently pending Promises in read/write error paths that caused polling hangs and forced disconnections
- Removes forced disconnect on polling overload that amplified create/destroy cycles triggering the segfault

## Root Cause Analysis

Three independent bugs formed a **failure cascade** on Linux:

1. **Bug 3 (instantiation)**: \snap7.S7Client()\ without \
ew\ created throwaway instances. When GC'd, the destructor called \constructor.Reset()\ on a static \Nan::Persistent<FunctionTemplate>\, corrupting the prototype for all subsequent instances → **segfault**.

2. **Broken Promise rejections**: \\_readVars\, \\_writeDB\, and \\_writeVars\ used \
eturn _getErr(err)\ instead of \
eturn reject(_getErr(err))\. The Promises were left permanently pending, the \working\ flag stayed \	rue\, the overloading counter accumulated, and the driver was forced into disconnect/reconnect cycles → **amplified segfaults**.

3. **Forced disconnect on overload**: When polling overloads accumulated (\overloading >= 3\), \\_checkWorking\ called \s7client.Disconnect()\, triggering device cleanup and re-instantiation → **further amplified segfaults**.

## Changes

| File | Change |
|------|--------|
| \server/runtime/devices/s7/index.js\ | \snap7.S7Client()\ → \
ew snap7.S7Client()\ (line 677) |
| \server/runtime/devices/s7/index.js\ | \_readVars\: \
eturn _getErr(err)\ → \
eturn reject(_getErr(err))\ |
| \server/runtime/devices/s7/index.js\ | \_writeDB\: \
eturn _getErr(err)\ → \
eturn reject(_getErr(err))\ |
| \server/runtime/devices/s7/index.js\ | \_writeVars\: \
eturn _getErr(err)\ → \
eturn reject(_getErr(err))\ |
| \server/runtime/devices/s7/index.js\ | \_checkWorking\: removed forced \s7client.Disconnect()\ on overload |

## Upstream Dependencies

This PR should be merged **after** the upstream fixes:

- **[davenardella/snap7#29](https://github.com/davenardella/snap7/pull/29)** — Fix POSIX thread joinable lifecycle in \snap_threads.cpp\ (C library)
- **[mathiask88/node-snap7#107](https://github.com/mathiask88/node-snap7/pull/107)** — Remove \constructor.Reset()\ in \~S7Client()\ and \~S7Server()\ destructors (Node.js wrapper)

While the upstream fixes address the crash at the native layer, these FUXA-side fixes are independently necessary because:
- Bug 3 (missing \
ew\) creates the throwaway instances that trigger the destructor path
- The Promise bugs cause operational instability regardless of the native crash
- The overload disconnect creates unnecessary create/destroy cycles

## Testing

- [x] Verified all three commits apply cleanly to \server/runtime/devices/s7/index.js\
- [x] Confirmed no other files are affected
- [x] Changes are backward-compatible (no API changes, no new dependencies)
- [x] Tested on Windows with Node 18 and VS2022 C++ build tools

### DIFF
--- a/server/runtime/devices/s7/index.js
+++ b/server/runtime/devices/s7/index.js
@@ -416,12 +416,7 @@ function S7client(_data, _logger, _events, _runtime) {
         if (check && working) {
             overloading++;
             logger.warn(`'${data.name}' working (connection || polling) overload! ${overloading}`);
-            // !The driver don't give the break connection
-            if (overloading >= 3) {
-                s7client.Disconnect();
-            } else {
-                return false;
-            }
+            return false;
         }
         working = check;
         overloading = 0;
--- a/server/runtime/devices/s7/index.js
+++ b/server/runtime/devices/s7/index.js
@@ -478,7 +478,7 @@ function S7client(_data, _logger, _events, _runtime) {
     var _readVars = function (vars) {
         return new Promise((resolve, reject) => {
             s7client.ReadMultiVars(vars, (err, res) => {
-                if (err) return _getErr(err);
+                if (err) return reject(_getErr(err));
                 let errs = [];
                 res = vars.map((v, i) => {
                     let value = null;
@@ -487,8 +487,7 @@ function S7client(_data, _logger, _events, _runtime) {
                     } else {
                         try {
                             if (v.type === 'BOOL') {
-                                // check the full byte and send all bit if there is a change
-                                value = datatypes['BYTE'].parser(res[i].Data);//, v.Start, -1);
+                                value = datatypes['BYTE'].parser(res[i].Data);
                             } else {
                                 value = datatypes[v.type].parser(res[i].Data);
                             }
@@ -525,7 +524,7 @@ function S7client(_data, _logger, _events, _runtime) {
             }
             let buffer = datatypes[v.type].formatter(v.value)
             s7client.DBWrite(DBNr, offset, end - offset, buffer, (err, res) => {
-                if (err) return _getErr(err);
+                if (err) return reject(_getErr(err));
                 resolve(changed);
             });
         });
@@ -553,7 +552,7 @@ function S7client(_data, _logger, _events, _runtime) {
         }));
         return new Promise((resolve, reject) => {
             s7client.WriteMultiVars(toWrite, (err, res) => {
-                if (err) return _getErr(err);
+                if (err) return reject(_getErr(err));
                 let errs = [];
 
                 res = vars.map((v, i) => {
--- a/server/runtime/devices/s7/index.js
+++ b/server/runtime/devices/s7/index.js
@@ -680,7 +680,7 @@ module.exports = {
     },
     create: function (data, logger, events, manager, runtime) {
         if (!loadSnap7Lib(manager)) return null;
-        datatypes = require('./datatypes')(snap7.S7Client ? snap7.S7Client() : snap7);
+        datatypes = require('./datatypes')(snap7.S7Client ? new snap7.S7Client() : snap7);
         return new S7client(data, logger, events, runtime);
     }
 }
--- a/server/runtime/devices/s7/index.js
+++ b/server/runtime/devices/s7/index.js
@@ -16,6 +16,9 @@ function S7client(_data, _logger, _events, _runtime) {
     var data = JSON.parse(JSON.stringify(_data));                   // Current Device data { id, name, tags, enabled, ... }
     var logger = _logger;               // Logger
     var s7client = new snap7.S7Client();// Client node-S7
+    if (!datatypes) {
+        datatypes = require('./datatypes')(s7client);
+    }
     var working = false;                // Working flag to manage overloading polling and connection
     var events = _events;               // Events to commit change to runtime
     var lastStatus = '';                // Last Device status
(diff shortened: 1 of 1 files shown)


unocelli:
Hi @JordiB-Inserma Thanks for the detailed RCA. The Promise rejection fixes look correct, and removing the forced disconnect on overload makes sense given the native create/destroy crash path.

Before merging, I’d like to clarify one point about the `datatypes` initialization:

```js
datatypes = require('./datatypes')(snap7.S7Client ? new snap7.S7Client() : snap7);
```
This now uses new, but it still creates a temporary S7Client only to read constants, and the instance is not retained afterwards. If the native crash is related to S7Client destruction / GC in node-snap7, could this temporary instance still trigger the problematic destructor path before the upstream node-snap7 fix is available?

Would it be possible to avoid this throwaway client, for example by initializing datatypes from the actual driver client instance or another stable constants object?

Also, since the issue is described as Linux-specific, could you confirm whether this branch has been tested on Linux, ideally with the referenced upstream snap7 / node-snap7 fixes applied?

JordiB-Inserma:
Hi @unocelli,

Thanks for the thorough review — your observation about the throwaway client is spot on, and it pushed us to dig deeper into the node-snap7 native layer. Here's what we found.

## The `new` fix is incomplete

After reading the node-snap7 C++ source (`src/node_snap7_client.cpp`), we confirmed that **the throwaway instance still triggers the problematic destructor path**. Adding `new` corrects a syntactic issue (calling a constructor without `new`), but does not address the core problem: a temporary `S7Client` is created, immediately loses all references, and gets garbage collected.

The destructor in node-snap7 does this:

```cpp
S7Client::~S7Client() {
  snap7Client->Disconnect();
  delete snap7Client;
  constructor.Reset();  // ← resets the SHARED Nan::Persistent<FunctionTemplate>
  uv_mutex_destroy(&mutex);
}
```

That `constructor.Reset()` call corrupts the shared prototype for all subsequent instances — this is the segfault root cause identified in [mathiask88/node-snap7#107](https://github.com/mathiask88/node-snap7/pull/107).

## Why we can't just pass the constructor

Our first instinct was to replace `new snap7.S7Client()` with `snap7.S7Client` (the constructor itself). But the node-snap7 constants (`S7WLBit`, `S7WLByte`, `S7WLWord`, etc.) are defined via `Nan::SetPrototypeTemplate` in the C++ source — meaning they exist **only on instances** (via the prototype), not as static properties on the constructor:

```cpp
// These are PROTOTYPE properties, NOT statics
Nan::SetPrototypeTemplate(tpl, "S7WLBit", Nan::New<v8::Integer>(S7WLBit), v8::ReadOnly);
Nan::SetPrototypeTemplate(tpl, "S7WLByte", Nan::New<v8::Integer>(S7WLByte), v8::ReadOnly);
// ...
```

So `snap7.S7Client.S7WLBit` returns `undefined` — we **need** an instance to read these values.

## Proposed solution: lazy init from the real client

Instead of creating a throwaway instance in `create()`, we move the `datatypes` initialization into the `S7client` constructor itself, using the actual `s7client` instance that will be retained for the connection lifetime:

**In `create()` (line 681-685):** remove the datatypes initialization:

```js
create: function (data, logger, events, manager, runtime) {
    if (!loadSnap7Lib(manager)) return null;
    return new S7client(data, logger, events, runtime);
}
```

**In `S7client()` constructor (after line 18):** initialize lazily with a guard:

```js
var s7client = new snap7.S7Client();
if (!datatypes) {
    datatypes = require('./datatypes')(s7client);
}
```

This way:
- No throwaway `S7Client` is ever created — zero risk of triggering the destructor path
- The real `s7client` instance is retained in the closure for the connection lifetime, so no GC/destructor issue
- The guard `if (!datatypes)` ensures it only runs once (first device initialization)
- The `datatypes` object contains only numeric constants (`0x01`, `0x02`, etc.) identical across all instances

## Regarding Linux testing

This branch was tested on Windows with Node 18. The Promise rejection fixes and overload disconnect removal are platform-independent correctness fixes. The `datatypes` lazy-init change we're proposing here is also platform-independent — it eliminates the throwaway instance entirely, so the destructor path is never reached regardless of platform or upstream fix status.

Linux testing is currently in progress with the upstream fixes from [mathiask88/node-snap7#107](https://github.com/mathiask88/node-snap7/pull/107) and [davenardella/snap7#29](https://github.com/davenardella/snap7/pull/29) applied.

## Changes applied

All four fixes are now implemented in this branch:

| Commit | Change | Status |
|--------|--------|--------|
| Lazy init datatypes | `create()` no longer creates throwaway `S7Client`; initialized lazily from the real instance in the constructor | ✅ Applied |
| Promise rejections | `_readVars`, `_writeDB`, `_writeVars`: `return _getErr(err)` → `return reject(_getErr(err))` | ✅ Applied |
| Overload disconnect | `_checkWorking`: removed forced `s7client.Disconnect()` on `overloading >= 3` | ✅ Applied |

The `new` keyword commit (d876199) is now superseded — the line it modified is removed entirely by the lazy init change.

unocelli:
Hi @JordiB-Inserma Good job, Thanks a lot!

PR #1784, opened 2025-06-02
feat/node-webcam

rikugun:
- Add support for node-webcam. when setTag with value  true, it'll take a snapshots with device settings.  
- Add a new action for refresh images.
 
## INSTALL:
1. follow https://www.npmjs.com/package/node-webcam ,install requires.
2. install plugins in fuxa

## USAGE:
image

image

image

image

PR #2444, opened 2026-07-18
Feature/reverse proxy

astrowild99:
## 📌 Description

Introducing a new ENV variable to deploy FUXA under a custom path (e.g. /fuxa) when working behind a reverse proxy.

The env variable is BASE_PATH and a working example is `BASE_PATH=/fuxa`.

The change keeps the current working structure, and is a runtime change, meaning that
you don't have to specify the path at compile time.

It has been tested both with docker behind an nginx proxy and on k8s with traefik ingress and
the Gateway API.

A tiny paragraph in the deploy documentation is therefore added, to explain how to work behind a reverse proxy.

> A few typo were fixed when working on the routing paths.

---

## 🧪 Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [x] Documentation
- [ ] Other

---

## 🚫 Build Artifacts Check

Please confirm:

---

## 🔍 Checklist

---

- [x] The documentation builds locally with `mkdocs serve`
- [x] All links were tested
- [x] Images use relative paths (e.g. `images/example.png`)
- [x] No references to the old GitHub Wiki

---


unocelli:
Hi @astrowild99 Thanks for working on reverse-proxy/base-path support. The direction makes sense, but I think you need to fix a few cases before this can be merged.

With `BASE_PATH=/fuxa`, some paths still look broken:

- The refresh cookie path is still hardcoded to `/api/refresh` in `server/api/auth/index.js`. Since the endpoint becomes `/fuxa/api/refresh`, the browser will not send the cookie and `enableRefreshCookieAuth` will break under a base path. Please update both `setRefreshCookie()` and `clearRefreshCookie()` to use the normalized base path.

- Snapshot image URLs are still generated as root-relative `/snapshots...` in `client/src/app/gauges/controls/html-image/html-image.component.ts`, while the server now mounts them at `BASE_PATH + '/snapshots'`. Please build these URLs through the same base-path logic used by API/socket URLs.

- `getListenPath()` currently prepends an extra `/` to `BASE_PATH`, so `BASE_PATH=/fuxa` logs `//fuxa/`.

- Rewriting `index.html`, JS and CSS files inside `settings.httpStatic` at startup is risky. It can fail on read-only deployments, and it also mutates the built assets permanently: starting once with `/fuxa` can leave the bundle rewritten even after restarting without `BASE_PATH`.

Please address these before merge, ideally with a manual/tested scenario for `BASE_PATH=/fuxa` covering API calls, socket connection, refresh-cookie auth, and snapshot images.

astrowild99:
Thank you for assistance! I will Address these issues within this week and submit an upgrade to this merge request.

astrowild99:
Hi @unocelli 
Once again thank you for the suggested edits, I should have addressed all the issues you pointed out. I opted to use a local system to cache html with my edit instead of changing the files on disk, thus being more resilient to both read-only filesystems and stale changes when the ENV is reverted. 
I couldn't find how to manually test the /snapshot endpoint, so I fixed it by sending the right address via code. If you could point to me in the docs where the /snapshot endpoint is used I can test it manually and be definitely sure. 
Thank you again!

unocelli:
Hi @astrowild99 Good job, Thanks a lot!

I noticed one remaining edge case: the Socket.IO connection now uses `location.origin`, while API calls still honor `environment.apiEndpoint` through `EndPointApi.getURL()`. That can regress deployments where the API/socket host is configured separately.

This is small and localized, so I’m fine handling it in a follow-up.

About snapshot:
I tested the latest version with the project I attached, without `BASE_PATH`, and I don't see regressions.
I also could not find specific documentation (#1784). You can use the attached project as a reference to test it manually, but you will need to install the `node-webcam` plugin first, and, of course, a USB webcam.

[webcam-test.json](https://github.com/user-attachments/files/30760173/webcam-test.json)

PR #2488, opened 2026-08-01
fix(demo): use bare tag ids for gauge variableId bindings

### LINKED ISSUE #2486: Demo project binds gauges with a variableId form nothing produces (`^~^`); HmiService.toVariableId() is unused
Following on from #2478 — same build, different finding. This one is source-level only, so please treat it as a question about intent rather than a bug report.

### What I noticed

`client/src/assets/project.demo.fuxap` binds its gauges with the separator form:

```json
"variableId": "Device A^~^Tag Byte"
```

but the tag it refers to is declared with a bare id:

```json
"tags": { "Tag Byte": { "id": "Tag Byte", "name": "Tag Byte", ... } }
```

and the runtime keys its variable map on whatever id arrives on the wire, verbatim:

```ts
// client/src/app/_services/hmi.service.ts:264
const originalId = message.values[idx].id;
updateVariable(originalId, value, timestamp, quality);
```

No server driver composes the separator form — `mqtt/index.js` and `s7/index.js` both emit `{ id: <device>, values: [{ id: <tag id>, ... }] }` with the bare tag id. So `Device A^~^Tag Byte` matches nothing that is ever published.

### Why it doesn't show up

The demo's device is `"enabled": false`, so it never connects and never publishes values. The mismatch is invisible in normal use of the demo.

### The part that made me raise it

`HmiService.toVariableId()` — the only thing that builds this form — appears to have **no callers**:

```ts
// client/src/app/_services/hmi.service.ts:654
public static toVariableId(src: string, name: string) {
    return src + HmiService.separator + name;
}
```

`grep -rn "toVariableId" client/src server` (excluding `node_modules`) returns just that definition. Which reads like the `^~^` form is legacy, the code that generated it is orphaned, and the shipped demo still carries it.

### Why it matters practically

The demo project is the natural reference for "how do I bind a gauge to a tag". I used it that way while building an MQTT-driven HMI, copied the `^~^` shape, and got a silent no-match — no console error, gauges simply never updated, SVG rendering fine. It took a while to stop suspecting my own device configuration. Anyone reading the demo as the runtime contract will land in the same place.

### Questions

1. Is `^~^` still intended for any path, or is it fully legacy now?
2. If legacy — would you like the demo's `variableId` values corrected to the bare tag ids, and/or `toVariableId()` removed?

Happy to send whichever you prefer; I didn't want to guess between updating the demo, deleting the helper, and documenting the contract. Also happy to be told I've misread it.

_Checked against `master` at 9fa97da (client-side reading only — I have not reproduced this against a running instance for the demo's disabled S7 device)._

sankalpsthakur:
### Summary
Demo project used legacy Device^~^Tag variableId form that never matches runtime bare tag ids. Corrected demo bindings to bare tag ids.

### Changes
- Updated all 13 `variableId` values in `client/src/assets/project.demo.fuxap`:
  - `Device A^~^Tag Byte` → `Tag Byte` (4 bindings)
  - `Device A^~^Tag Bool` → `Tag Bool` (9 bindings)
- Left `variable` / `variableSrc` and `HmiService.toVariableId()` unchanged (demo-only, high merge odds).

### Validation
- inspected demo JSON / grep for remaining `^~^` in demo project (0 remaining)
- parsed demo project as JSON; all 13 `variableId` values are bare declared tag ids (`Tag Byte`, `Tag Bool`) matching `devices["Device A"].tags`

Fixes #2486

### AI/LLM disclosure
- AI coding tools (including Grok and/or Codex agent-assisted editing) were used to help draft or modify code and this PR description.
- I reviewed the complete change, understand the reasoning, and ran the reported local tests before submitting.
- This submission is original work of authorship under the project contributor terms; AI output was not pasted unreviewed.


unocelli:
Hi @sankalpsthakur, It was an old format, Thanks a lot!

PR #2518, opened 2026-08-20
feat: configure MQTT retain flag per topic

Samyak2406:
## Summary

- Add a retain checkbox to the MQTT publish-topic editor.
- Persist and reload the retain setting as a boolean in tag options.
- Apply the configured retain flag to JSON, raw, payload, echo, and legacy publish paths.
- Preserve retain=true for missing or non-boolean configuration.
- Add the retain label to every shipped locale.

## Validation

- 
- All 13 locale JSON files parse and contain the retain label.
- No-network smoke validation: 13/13 tests pass with the solution; the pristine control has the six expected feature failures.
- No dependency manifests changed.

Client Angular/Karma tests and production build were not run because client dependencies are unavailable in the current environment.


unocelli:
Hi @Samyak2406 Good job, Thanks a lot!

PR #2520, opened 2026-08-21
fix: align QuestDB addDaqValues guard with sqlite implementation

VeikkoAJ:
## 📌 Description

Restoring values when using QuestDB did not work because of a bug in the `addDaqValues`  guard. Only tags with invalid values were added to the restore array:
```js
 if (!tag.daq || utils.isNullOrUndefined(tag.value) || Number.isNaN(tag.value)) {
                if (tag.daq && tag.daq.restored) {
``` 
This resulted in tags with restore enabled being restored to default value or value in project data instead of last value in database.

Changed it to work in the same way as the sqlite implementation.

### How to replicate
1. Setup QuestDB.
2. Add new tag and configure to be restored
3. Change the tag value in UI so that it gets written to the database
4. Open tag options and press OK without changes.
5. Tag value reverts back to stale data

---

## 🧪 Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Other

---

## 🚫 Build Artifacts Check

Please confirm:

---

## 🔍 Checklist

---

## 📝 Additional Notes

### DIFF
--- a/server/runtime/storage/questdb/index.js
+++ b/server/runtime/storage/questdb/index.js
@@ -37,13 +37,14 @@ function QuestDB(_settings, _log, _currentStorage) {
 
         for (const tagid in tagsValues) {
             let tag = tagsValues[tagid];
-            if (!tag.daq || utils.isNullOrUndefined(tag.value) || Number.isNaN(tag.value)) {
-                if (tag.daq && tag.daq.restored) {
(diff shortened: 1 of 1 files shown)


unocelli:
Hi @VeikkoAJ Good job, Thanks a lot!

Issue #2485, opened 2026-07-31
[BUG] S7 device silently skipped when optional node-snap7 is missing


JCL-ingenia:
**FUXA 1.3.3**

The S7 adapter loads `node-snap7` as an optional dependency. When the module is not installed, the device is simply not started and `plugin is missing` appears only in the server log — the UI lets you create and enable the S7 connection with no visible error, which sends users down the wrong debugging path (network, rack/slot, firewall...) while the driver was never running. Related context: #2179.

**Suggestion**: surface the missing-plugin condition in the UI (device status = "driver not installed"), and/or fail loudly when an enabled device has no loadable driver.

PR #2549, opened 2026-09-20
Report missing device plugins as failed status

RevanthMeda:
Summary

- Keep track of enabled devices that cannot be instantiated because their plugin/driver is unavailable.
- Expose those devices through the existing "connect-failed" status instead of silently omitting them.
- Emit the normal "device-status:changed" event so the UI can show the failed state immediately.
- Clear the failed state when a device loads successfully or is removed.
- Add regression coverage for enabled and disabled devices.

This addresses #2485.

The existing server warning remains in place, while the configured device is now also visible to clients as failed instead of disappearing from the device status response.

Testing

Added "server/test/devices/missingPluginStatus.test.js" covering:

- An enabled device with an unavailable plugin reports "connect-failed" and emits a status event.
- A disabled device is not reported as failed.

The regression tests were added as part of this change. They have not been executed locally in this GitHub-only workflow.


RevanthMeda:
@unocelli When you have a chance, could you review this change? It addresses #2485 by keeping devices with unavailable plugins visible through the existing "connect-failed" status instead of silently omitting them. I kept the change focused and added regression coverage.

PR #2550, opened 2026-09-20
Fix uploaded resource URLs behind BASE_PATH

### LINKED ISSUE #2533: [BUG] /api/upload returns image location without BASE_PATH prefix — uploaded menu/header images show 404 behind reverse proxy
**Describe the bug**

When FUXA runs behind a reverse proxy with the `BASE_PATH` environment variable
(introduced in #2444), any image uploaded through the UI (e.g. Layout settings →
Navigation Side Menu → Menu Item → *My Image*) uploads successfully but never
displays. The sidenav shows an empty/broken image.

Root cause: `POST /api/upload` returns a hard-coded root-absolute location that
does not include `BASE_PATH`:

```js
// server/api/projects/index.js (line ~284)
let result = {'location': '/' + runtime.settings.httpUploadFileStatic + '/' + relativePath };
// → "/resources/<filename>"
```

The static route, however, is mounted **with** the prefix:

```js
// server/main.js
app.use(BASE_PATH + '/' + settings.httpUploadFileStatic, express.static(settings.uploadFileDir));
// → served at /fuxa1/resources/<filename>
```

The stored `"/resources/<filename>"` is saved into
`hmi.layout.navigation.items[].image` and rendered as ``.
Since a root-absolute URL ignores the injected `<base href="/fuxa1/">`, the
browser requests it at the domain root, which the reverse proxy (or FUXA itself)
answers with 404.

**Without** `BASE_PATH` (default standalone setup) everything works fine: the
static route is mounted at `/resources` and the returned `"/resources/<filename>"`
matches it exactly. The bug only manifests when `BASE_PATH` is set — which is
likely why it has gone unnoticed since #2444.

Verified against my instance (BASE_PATH=/fuxa1, nginx in front):

```
GET /resources/A-8055_icon.png        → 404   (path stored by FUXA)
GET /fuxa1/resources/A-8055_icon.png  → 200   (actual location)
```

**To Reproduce**

1. Start FUXA with `BASE_PATH=/fuxa1` behind any reverse proxy that maps
   `/fuxa1/` to the FUXA port (setup as documented in #2444).
2. Editor → Layout settings → Navigation Side Menu → add/edit a Menu Item.
3. In the icon dropdown choose *My Image* and upload any PNG.
4. Save, reload the runtime page, open the side menu.

**Expected behavior**

The uploaded image is displayed in the side menu.

**Actual behavior**

Empty space where the image should be; DevTools shows the request for
`/resources/<filename>` returning 404. The same happens for header items and
any other place that consumes the `location` returned by `/api/upload`.

**Suggested fix**

Return the location without the leading slash so it resolves through the
injected `<base href>` (works both with and without BASE_PATH, since base href
is `/` in the default setup):

```js
let result = {'location': runtime.settings.httpUploadFileStatic + '/' + relativePath };
```

or alternatively prepend the base path explicitly:

```js
const basePath = process.env.BASE_PATH || '';
let result = {'location': basePath + '/' + runtime.settings.httpUploadFileStatic + '/' + relativePath };
```

Existing projects that already stored `/resources/...` paths would still need a
one-time migration (strip the leading slash), which can be done via
`POST /api/projectData` `cmd:'layout'`.

**Environment**

- FUXA v1.3.4 (also present on current `master`, `server/api/projects/index.js` line 284)
- Node.js server on Windows 10, nginx reverse proxy, `BASE_PATH=/fuxa1`
- Related: #2444 (BASE_PATH feature), #2526 (another BASE_PATH integration issue)

RevanthMeda:
Summary

Fix uploaded resource URLs when FUXA is deployed below a "BASE_PATH", for example "/fuxa1/".

"POST /api/upload" currently returns a root-absolute URL such as:

"/resources/menu-icon.svg"

A root-absolute URL bypasses the application's configured base path, causing uploaded menu/header images and other uploaded resources to return 404 when FUXA is hosted behind a reverse proxy on a subpath.

This change returns a browser-relative resource URL instead:

"resources/menu-icon.svg"

The browser can then resolve it through FUXA's configured "<base href>", while standalone deployments continue to resolve it from "/".

Fixes #2533.

Changes

- Remove the leading "/" from uploaded resource locations.
- Add an API regression test verifying that:
  - the upload succeeds;
  - the returned location is "resources/menu-icon.svg";
  - the returned URL is not root-absolute;
  - the resource is actually written to the configured upload directory.

Testing

Added:

"server/test/api/uploadBasePath.test.js"

The test could not be executed in my current environment because outbound GitHub/network access is unavailable. The change and committed branch diff were verified separately.


RevanthMeda:
@unocelli @astrowild99 When you have a chance, could you review this change? It fixes an upload-resource path edge case related to the BASE_PATH support from #2444 by returning a browser-relative resource URL instead of a root-absolute path.

